PR #1, opened 2025-07-24
setup build/lint system. add rms_norm as first op to verify functionality

jikunshang:
build/lint system are mostly copied from vllm project.


tests/test_layernorm.py
```diff
@@ -0,0 +1,153 @@
...
+
+@pytest.mark.parametrize("num_tokens", NUM_TOKENS)
+@pytest.mark.parametrize("hidden_size", HIDDEN_SIZES)
+@pytest.mark.parametrize("add_residual", ADD_RESIDUAL)
+@pytest.mark.parametrize("dtype", DTYPES)
+@pytest.mark.parametrize("seed", SEEDS)
+@pytest.mark.parametrize("device", XPU_DEVICES)
+@pytest.mark.parametrize("strided_input", [False, True])
+@torch.inference_mode()
+def test_rms_norm(
```
dbyoung18, line 31: Do we still need another UTs to track XPU kernels or just reuse UTs under vllm? 
https://github.com/vllm-project/vllm/blob/main/tests/kernels/core/test_layernorm.py#L31

tests/ops.py
```diff
@@ -0,0 +1,199 @@
...
+import logging
+from typing import Optional, Union
+
+import torch
+import torch.nn as nn
+
+logger = logging.getLogger("vllm_xpu_kernel")
+
+
+class CustomOp(nn.Module):
```
dbyoung18, line 13: This part also looks like a duplicate from https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/custom_op.py#L15, can we reuse vllm?

requirements.txt
```diff
@@ -0,0 +1,19 @@
...
+# build
+regex 
+cmake>=3.26
+packaging>=24.2
+setuptools-scm>=8
+setuptools>=77.0.3,<80.0.0
+wheel
+
+# torch dependency
+torch @ https://ubit-artifactory-ba.intel.com/artifactory/aipc_releases-ba-local/gpu-new/releases/2025.1/IPEX_2.8.10+xpu/RC2/vllm_pytorch_whl_310/torch-2.8.0+xpu-cp310-cp310-linux_x86_64.whl
```
dbyoung18, line 13: internal link here as a placeholder?

tests/ops.py
jikunshang: I prefer not reuse vllm, because it may introduce potential loop import issue

.gitignore
```diff
@@ -0,0 +1,202 @@
+# version file generated by setuptools-scm
```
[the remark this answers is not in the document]
jikunshang: yes, it's copied from vllm. I prefer to keep it same with vllm's

tests/ops.py
xuechendi: How about register using vllm.general_plugins
```
    entry_points={
        "vllm.general_plugins": ["xpu_custom_ops = vllm_xpu_kernels:register_ops"],
    },
```

jikunshang: using vllm plugin is a good idea. but I feel we don't need register_ops like what guadi side do. I will take a look.

jikunshang: double checked plugin system, it will make vllm as dependency which is not what we want. besides, this is only for unit test, so we can just keep it.

PR #4, opened 2025-07-31
refactor files by ops in test folder

jikunshang:
refactor `custom_ops.py`, keep interface only , move `RmsNorm` to standalone file.

### DIFF
--- a/benchmark/benchmark_layernorm.py
+++ b/benchmark/benchmark_layernorm.py
@@ -6,7 +6,7 @@ from argparse import ArgumentParser
 
 import torch
 
-from tests.ops import RMSNorm
+from tests.ops.layernorm_op import RMSNorm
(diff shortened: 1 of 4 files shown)


jikunshang:
> What will the usage of vllm-xpu-kernels be like in the upstream vllm? From my view, it's good to refactor the dispatch progress and make it look as similar as upstream's usage.

we need do such code change on vllm upstream side:
1. import this library (only need once)
2. for each op, we need modify forward_xpu method, switch from ipex impl to kernel impl. and this should be same to cuda. we can copy forward_cuda or just call `forward_cuda` if possible. [ eg](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/layernorm.py#L173-L195)

baodii (approved):
LGTM

PR #12, opened 2025-08-13
add fp8 quantization kernels 

baodii:
add a) static_scaled_fp8_quant; b) dynamic_scaled_fp8_quant; c) dynamic_per_token_scaled_fp8_quant kernels.

### DIFF
--- a/csrc/xpu/quantization/fp8/fp8_quant.cpp
+++ b/csrc/xpu/quantization/fp8/fp8_quant.cpp
@@ -9,6 +9,7 @@
 
 #include "fp8_quant.h"
 #include "quant_utils.h"
+#include "xpu/utils.h"
 
 namespace vllm {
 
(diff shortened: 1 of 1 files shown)


csrc/xpu/quantization/fp8/fp8_quant.cpp
```diff
@@ -0,0 +1,224 @@
...
+                             torch::Tensor const& scale)  // [1]
+{
+  int64_t num_tokens = input.numel() / input.size(-1);
+  int64_t num_elems = input.numel();
+  sycl::range<1> grid(num_tokens);
+  sycl::range<1> block(1024);
+  at::Device curDevice = at::Device(at::kXPU, at::xpu::current_device());
+  at::DeviceGuard device_guard(curDevice);
+
+  auto stream = at::xpu::getCurrentXPUStream().queue();
```
jikunshang, line 132: maybe can just call `auto& queue = vllm::xpu::vllmGetQueue()`

jikunshang (approved):
LGTM!

csrc/xpu/quantization/fp8/fp8_quant.cpp
rogerxfeng8: kernel should not depend on vllm api or libs

PR #13, opened 2025-08-14
Onednn w8a16gemm [WIP]

zufangzhu:
Migrating onednn w8a16 gemm

working on benchmarks and tests..


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "third_party/oneDNN"]
```
jikunshang, line 1: do we need some extra steps for git submodule in build steps?

csrc/xpu/torch_bindings.cpp
```diff
@@ -63,6 +63,12 @@ TORCH_LIBRARY_EXPAND(TORCH_EXTENSION_NAME, ops) {
       "()");
   ops.impl("dynamic_per_token_scaled_fp8_quant", torch::kXPU,
            &dynamic_per_token_scaled_fp8_quant);
+
+  // w8a16 onednn gemm
+  ops.def(
```
jikunshang, line 68: for such kernel which don't have same interface in vllm-cuda, maybe we'd better move to `_xpu_C` extension like what `_rocm_C` extension did. cc @dbyoung18

dbyoung18: Agree to follow vllm's design, isolate xpu-only kernels to _xpu_C.
I have a similar case of `deepseek_rope`, will refactor cmake and submit a PR once ready for u to refer.

CMakeLists.txt
```diff
@@ -146,12 +149,12 @@ endif()
...
-    "csrc/xpu/activation.cpp"
-    "csrc/xpu/pos_encoding_kernels.cpp"
-    "csrc/xpu/torch_bindings.cpp"
-    "csrc/xpu/quantization/fp8/fp8_quant.cpp"
+    "csrc/cache.cpp"
+    "csrc/layernorm.cpp"
+    "csrc/activation.cpp"
+    "csrc/pos_encoding_kernels.cpp"
+    "csrc/torch_bindings.cpp"
+    "csrc/quantization/fp8/fp8_quant.cpp"
```
dbyoung18, line 157: agree refactor the structure a little bit here, make it more like the upstream vllm

CMakeLists.txt
```diff
@@ -176,6 +179,45 @@ define_gpu_extension_target(
...
+  set(CMAKE_CXX_COMPILER icpx)
+  set(VLLM_EXTRA_INCLUDE_DIRECTORIES ${CMPLR_ROOT}/include/sycl)
+  list(APPEND VLLM_GPU_FLAGS "-DVLLM_BUILD_XPU_OPS" )
+  list(APPEND VLLM_GPU_LINK_FLAGS "-fsycl" "-fsycl-targets=spir64")
+  list(APPEND VLLM_LINK_LIBRARIES "sycl" "OpenCL" "pthread" "m" "dl" "torch" )
+endif()
+
+if(ONEDNN_FOUND)
+  list(APPEND VLLM_EXT_XPU_SRC
+    "csrc/xpu/onednn/fp8_gemm_w8a16.cpp"
```
dbyoung18, line 200: Good refactor! As for future, we may have xpu special sycl/onednn/cutlass kernels under `_xpu_C`, better to add subfolders to isolate each like what u do here.

tests/test_fp8_linear.py
```diff
@@ -0,0 +1,54 @@
+# SPDX-License-Identifier: Apache-2.0
```
dbyoung18, line 1: +@jikunshang
Looks this part is necessary for upstream. If so, I would suggest isolate this part outside `tests`. `tests` should put xpu ut related codes maintained by ourselves.
For upstream related changes besides c++ code in .so, we may need to consider a right place to put and use.

dbyoung18 (commented):
Main part LGTM.
Since it's the first PR for _xpu_C and onednn. May need more review comments here.

PR #14, opened 2025-08-18
use function instead of lambda

zufangzhu:
remove lambda in sycl kernel submit

### DIFF
--- a/csrc/xpu/cache.cpp
+++ b/csrc/xpu/cache.cpp
@@ -11,61 +11,90 @@
 namespace vllm {
 
 template <typename scalar_t, typename cache_t, Fp8KVCacheDataType kv_dt>
-void reshape_and_cache_kernel(
-    const scalar_t* __restrict__ key,    // [num_tokens, num_heads, head_size]
-    const scalar_t* __restrict__ value,  // [num_tokens, num_heads, head_size]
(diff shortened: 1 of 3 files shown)


zufangzhu:
> The lambda -> functor LGTM.
> 
> BTW, another trivial suggestion. [`vllm-xpu-kernels/tests/ops`](https://github.com/vllm-project/vllm-xpu-kernels/tree/main/tests/ops) is mapped w/ [`vllm/model_executor/layers`](https://github.com/vllm-project/vllm/tree/main/vllm/model_executor/layers/), designed for the ops need other function warp or abstract to `class xxx(nn.Module)`. There is no need for [cache_ops.py](https://github.com/vllm-project/vllm-xpu-kernels/blob/main/tests/ops/cache_op.py) here, just use `from tests.register_ops import reshape_and_cache, reshape_and_cache_flash` in https://github.com/vllm-project/vllm-xpu-kernels/blob/main/tests/test_cache.py#L8 should be enough and more similar to upstream usage.

Thank you for valuable suggestions! Updates have been made in this PR.

jikunshang (approved):
LGTM!

jikunshang:
will functor/ lambda influence performance?

zufangzhu:
> will functor/ lambda influence performance?

Nope, I run some benchmarks on PVC, it shows same perf as before~

PR #16, opened 2025-08-19
Add activation: gelu_fast/gelu_new/gelu_quick

zhenwei-intel:
reference:
- https://github.com/vllm-project/vllm/blob/main/tests/kernels/core/test_activation.py
- https://github.com/vllm-project/vllm/blob/main/csrc/activation_kernels.cu

Currently known issues:
- `gelu_new_kernel` fails the accuracy test when using `sycl::tanh`, currently using `cmath's tanhf` instead.

### DIFF
--- a/csrc/xpu/activation.cpp
+++ b/csrc/xpu/activation.cpp
@@ -31,7 +31,7 @@ inline T gelu_new_kernel(const T& x) {
 template <typename T>
 inline T gelu_quick_kernel(const T& x) {
   // x * sigmoid(1.702 * x)
-  return (T)(((float)x) / (1.0f + sycl::native::exp(-1.702f * (float)x)));
+  return (T)(((float)x) / (1.0f + (T)sycl::exp(-1.702f * (float)x)));
 }
 
 template <typename scalar_t, scalar_t (*ACT_FN)(const scalar_t&),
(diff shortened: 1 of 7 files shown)


jikunshang:
Please use functor instead of lambda. can refer https://github.com/vllm-project/vllm-xpu-kernels/pull/15

csrc/xpu/activation.cpp
```diff
@@ -12,12 +12,69 @@ inline T silu_kernel(const T& x) {
...
+template <typename T>
+inline T gelu_new_kernel(const T& x) {
+  // 0.5 * x * (1.0 + tanh(0.7978845608 * (x + 0.044715 * x * x * x)))
+  const float x3 = (float)(x * x * x);
+  const T t =
+      (T)sycl::tanh((T)(0.79788456f * (float)(x + (T)(0.044715f * x3))));
+  return ((T)0.5) * x * (((T)1.0) + t);
+}
+
+using sycl_bf16 = sycl::ext::oneapi::bfloat16;
```
jikunshang, line 32: I prefer use `sycl_bf16_t`

jikunshang (approved):
LGTM.

csrc/xpu/activation.cpp
zhenwei-intel: updated

PR #17, opened 2025-08-19
feat(moe_sum): add moe_sum

dbyoung18:
* refactor cmake for moe related kernels targeting included in `_moe_C`
* add sycl kernel `moe_sum`
* verified accuracy pass


mayuyuace:
Will register.py  be moved to other directory instead of tests?

baodii (approved):
LGTM

PR #20, opened 2025-08-21
refactor the structure

zufangzhu:
under csrc/, the structure as below:

image


csrc/xpu/ops.h
```diff
@@ -1,38 +1,3 @@
 #pragma once
 
-#include <torch/all.h>
```
jikunshang, line 3: maybe we can remove these two files and add in next PR for these kernels?

jikunshang (approved):

baodii (approved):
LGTM

PR #22, opened 2025-08-22
[WIP] Grouped gemm cutlass

Liangliang-Ma:
developing grouped_gemm_fp8 for deepseek fused moe.
based on https://github.com/vllm-project/vllm-xpu-kernels/pull/11 cutlass env.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,16 +147,64 @@ endif()
 
 if(VLLM_GPU_LANG STREQUAL "SYCL")
   set(VLLM_EXT_SRC
+    "csrc/xpu/cutlass_sycl_demo.cpp"
     "csrc/xpu/layernorm.cpp"
     "csrc/xpu/torch_bindings.cpp"
   )
   include_directories("/usr/include")
-  set(CMPLR_ROOT $ENV{CMPLR_ROOT})
-  set(CMAKE_CXX_COMPILER icpx)
+  list(APPEND VLLM_INCLUDE_DIR ${CMPLR_ROOT}/include/)
+  list(APPEND VLLM_INCLUDE_DIR ${CMPLR_ROOT}/include/sycl/)
+  list(APPEND VLLM_INCLUDE_DIR ${CMPLR_ROOT}/include/sycl/syclcompat/)
+  message(STATUS "VLLM_INCLUDE_DIR: ${VLLM_INCLUDE_DIR}")
   set(VLLM_EXTRA_INCLUDE_DIRECTORIES ${CMPLR_ROOT}/include/sycl)
   list(APPEND VLLM_GPU_FLAGS "-DVLLM_BUILD_XPU_OPS" )
-  list(APPEND VLLM_GPU_LINK_FLAGS "-fsycl" "-fsycl-targets=spir64")
+  list(APPEND VLLM_GPU_LINK_FLAGS "-fsycl" "-fsycl-targets=spir64" "-Xspirv-translator" "-spirv-ext=+SPV_INTEL_split_barrier")
   list(APPEND VLLM_LINK_LIBRARIES "sycl" "OpenCL" "pthread" "m" "dl" "torch" )
+
+
+  # add cutlass dependency
+  set(CUTLASS_ENABLE_HEADERS_ONLY "ON" CACHE BOOL "Enable only the header library")
+
+  # Set CUTLASS_REVISION. Used for FetchContent. Also fixes some bogus messages when building.
+  set(CUTLASS_REVISION "v3.9-0.2" CACHE STRING "CUTLASS revision to use")
+
+  # Use the specified CUTLASS source directory for compilation if VLLM_CUTLASS_SRC_DIR is provided
+  FetchContent_Declare(
+      cutlass-sycl
+      GIT_REPOSITORY https://github.com/codeplaysoftware/cutlass-sycl.git
+      # Please keep this in sync with CUTLASS_REVISION line above.
+      GIT_TAG ${CUTLASS_REVISION}
+      GIT_PROGRESS TRUE
+
+      # Speed up CUTLASS download by retrieving only the specified GIT_TAG instead of the history.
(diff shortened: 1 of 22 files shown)


CMakeLists.txt
```diff
@@ -171,12 +171,13 @@ if(VLLM_GPU_LANG STREQUAL "SYCL")
...
 
   # Set CUTLASS_REVISION. Used for FetchContent. Also fixes some bogus messages when building.
-  set(CUTLASS_REVISION "main" CACHE STRING "CUTLASS revision to use")
+  set(CUTLASS_REVISION "dev" CACHE STRING "CUTLASS revision to use")
 
   # Use the specified CUTLASS source directory for compilation if VLLM_CUTLASS_SRC_DIR is provided
   FetchContent_Declare(
       cutlass-sycl
-      GIT_REPOSITORY https://github.com/intel/cutlass-sycl
+      GIT_REPOSITORY https://github.com/Liangliang-Ma/cutlass-sycl
```
rogerxfeng8, line 179: why using private forked cutlass-sycl?

Liangliang-Ma: I will rebase to cutlass-sycl/main.

csrc/xpu/cutlass_kernels/grouped_gemm_kernel.cpp
```diff
@@ -0,0 +1,471 @@
...
+      stride_A_host.push_back(
+          cutlass::make_cute_packed_stride(StrideA{}, {M, K, 1}));
+      stride_B_host.push_back(
+          cutlass::make_cute_packed_stride(StrideB{}, {N, K, 1}));
+      stride_C_host.push_back(
+          cutlass::make_cute_packed_stride(StrideC{}, {M, N, 1}));
+      stride_D_host.push_back(
+          cutlass::make_cute_packed_stride(StrideD{}, {M, N, 1}));
+    }
+  }
```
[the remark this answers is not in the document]
Liangliang-Ma: The M dimension will be handled in another kernel and current D2H is supper light weighted.

vllm_xpu_kernels/fused_moe_interface.py
```diff
@@ -0,0 +1,183 @@
...
+def cutlass_grouped_gemm(input_A, input_B, output, offset, n, k, num_experts):
+    alpha = torch.ones(num_experts, dtype=torch.float32, device=input_A.device)
+    beta = torch.zeros(num_experts, dtype=torch.float32, device=input_A.device)
+    gemm_args = prepare_gemm_args(n, k, offset, input_A, input_B, output,
+                                  alpha, beta, num_experts)
+    offset = torch.tensor(offset, dtype=torch.int64, device="cpu")
+    torch.ops._xpu_C.cutlass_grouped_gemm(offset=offset, N=n, K=k, **gemm_args)
+
+
+def cutlass_fused_moe(hidden_states, w13, w2, topk_weights, topk_ids,
```
jikunshang, line 80: I think this may not be a good interface... it's ok to keep it for now.

jikunshang (approved):
some minor comments. can address in next PR.

Liangliang-Ma:
> some minor comments. can address in next PR.

Got it. Thx

csrc/xpu/cutlass_kernels/grouped_gemm_kernel.cpp
```diff
@@ -0,0 +1,471 @@
...
+  using EpilogueDispatchPolicy = cutlass::epilogue::IntelXeXMX16Group;
+  using EpilogueOp =
+      cutlass::epilogue::fusion::LinearCombination<float_t, float_t>;
+
+  using CollectiveEpilogue =
+      typename cutlass::epilogue::collective::CollectiveBuilder<
+          cutlass::arch::IntelXe, cutlass::arch::OpClassTensorOp, TileShape,
+          Shape<_1, _1, _1>, cutlass::epilogue::collective::EpilogueTileAuto,
+          float, float, float, LayoutC, 1, ElementOutput, LayoutC, 1,
+          EpilogueDispatchPolicy, EpilogueOp>::CollectiveOp;
```
sanchitintel, line 443: Based on `xe_builder.cpp` & this code, it seems you used https://github.com/intel/cutlass-sycl/pull/505 as reference. 
Currently, it's using `EpilogueBuilder`, but I'll replace that code with `CollectiveEpilogue`, which is more configurable.

Thanks!

vllm_xpu_kernels/fused_moe_interface.py
```diff
@@ -0,0 +1,183 @@
...
+        exp_token_idxs = token_idxs[start_idx:end_idx]
+        # expert_tokens = hidden_states[exp_token_idxs]
+        # grouped_input_A.append(expert_tokens)
+        input_A[start_idx:end_idx, :].copy_(hidden_states[exp_token_idxs])
+
+    while len(offset) < num_experts:
+        offset.append(0)
+
+    ########### gemm1 ##################
+    input_B = w13.transpose(-1, -2).contiguous().transpose(-1, -2)
```
sanchitintel, line 143: You can use `ColumnMajor` `B` in the GroupGEMM kernel, so that you wouldn't have to transpose `B`.
To use column-major `B` (to avoid transposing weights), you can use a different copy atom for transposed loads . Coincidentally, because of how thread-value assignment works in the copy-atoms, the transpose copy atom for 16-bit (or 8-bits, for that matter) dtypes will load data in VNNI format (which is also true for atoms ending in `_N`).

However, if the latency of `tranpose of B` + `GEMM with RowMajor B` is lower than `GEMM with columnMajor B`, then you might want to retain this approach.

FWIW, once 32x32 transpose copy-atom for BF16 is added in cutlass, perf of `GEMM with columnMajor B` will become a bit better.

Thanks!

csrc/xpu/cutlass_kernels/grouped_gemm_kernel.cpp
sanchitintel: I explicitly attributed the reference in the description of https://github.com/intel/cutlass-sycl/pull/505.
Not only does it give credit to the original author, but it makes maintenance easier.

PR #36, opened 2025-09-15
reduce the kernel test coverage for simulator

chaojun-zhang:
## Purpose
reduce the kernel test coverage for simulator

design:

1. Use XPU_KERNEL_PYTEST_PROFILER to select the pytest profiler.（e.g. export XPU_KERNEL_PYTEST_PROFILER =MINI)

2. Based on the selected profiler, each UT file uses {pytest_profiler} to override pytest parameters, thereby controlling shapes under the simulator and reducing test time.
3. Then you can add code similar to the following in your test python file:

```python
MINI_PYTEST_PARAMS = {
    "default": {
        "num_tokens": [1],
        "head_size": [64, 80],
    },
    "test_concat_and_cache_mla": {
        "num_tokens": [1],
        "num_blocks": [32],
    },
    "test_gather_cache_mla": {
        "num_blocks": [32],
        "max_seq_len": [64],
    },
}
```

For example, **test_concat_and_cache_mla** here is the name of a unit test function, and the corresponding dictionary contains the shapes that need to be controlled .

- If no corresponding method is found, the **default** entry will be used by default, and if the **default** is also not defined, it means there is no shapes control .
- If you don't want to include any simulator shape parameters, you can ignore this  **MINI_PYTEST_PARAMS** configuration .

## Test Plan
XPU_KERNEL_PYTEST_PROFILER=MINI pytest -s -v  tests
or 
export XPU_KERNEL_PYTEST_PROFILER=MINI 
pytest -s -v  tests

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,13 +3,13 @@ import os
 
 
 def pytest_generate_tests(metafunc):
-    use_simulator = os.getenv("USE_SIMULATOR", "0") == "1"
-    if not use_simulator:
+    use_mini_pytest_profile = os.getenv("PYTEST_PROFILE", "") == "MINI"
+    if not use_mini_pytest_profile:
         return
 
     module = metafunc.module
 
-    func_pytest_params = getattr(module, "SIMULATOR_PYTEST_PARAMS", {})
+    func_pytest_params = getattr(module, "MINI_PYTEST_PARAMS", {})
     profile = func_pytest_params.get(metafunc.function.__name__, None)
 
     if not profile:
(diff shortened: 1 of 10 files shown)


jikunshang:
@wendyliu235 Please take a look.

rogerxfeng8:
> I feel we should avoid use `simulator`, maybe use something like `MINIMAL_PYTEST_PARAMS`. cc @rogerxfeng8

Agree. MINI_PYTEST_PARAMS (MINI, FULL) look more general names.

jikunshang (approved):

rogerxfeng8 (approved):

PR #39, opened 2025-09-17
topk softmax

mayuyuace:
qwen next need experts=512 kernel.

### DIFF
--- a/csrc/moe/moe_ops.h
+++ b/csrc/moe/moe_ops.h
@@ -15,3 +15,7 @@ std::tuple<torch::Tensor, torch::Tensor> fused_grouped_topk(
     const int64_t n_topk_group, const c10::string_view scoring_func,
     const double routed_scaling_factor,
     const c10::optional<torch::Tensor>& bias);
+
+void topk_softmax(torch::Tensor& topk_weights, torch::Tensor& topk_indices,
+                  torch::Tensor& token_expert_indices,
+                  torch::Tensor& gating_output);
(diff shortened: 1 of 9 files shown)


tests/test_topk_softmax.py
```diff
@@ -0,0 +1,50 @@
...
+        "n_token": [1],
+        "n_expert": [16, 512],
+    },
+}
+
+
+@pytest.mark.parametrize("n_token", [1, 33, 64])
+@pytest.mark.parametrize("n_hidden", [1024])
+@pytest.mark.parametrize("n_expert", [16, 192, 512, 1024])
+@pytest.mark.parametrize("topk", [2])
```
jikunshang, line 20: maybe we should add more topk? gpt oss is 4, deepseek r1 is 8

csrc/moe/moe_ops.h
```diff
@@ -15,3 +15,7 @@ std::tuple<torch::Tensor, torch::Tensor> fused_grouped_topk(
     const int64_t n_topk_group, const c10::string_view scoring_func,
     const double routed_scaling_factor,
     const c10::optional<torch::Tensor>& bias);
+
+void topk_softmax(torch::Tensor& topk_weights, torch::Tensor& topk_indices,
+                  torch::Tensor& token_expert_indices,
+                  torch::Tensor& gating_output, const bool renormalize);
```
jikunshang, line 21: vllm cuda doesn't have renormalize
https://github.com/vllm-project/vllm/blob/main/csrc/moe/topk_softmax_kernels.cu#L521-L525
any reason we added this in ipex?

PR #47, opened 2025-09-25
[UT][GroupedGemm] Add grouped gemm pytest mini scope

Liangliang-Ma:
1. add grouped gemm pytest mini scope
2. make confest be able to handle grouped param, eg "m,n,k"

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -20,7 +20,8 @@ def pytest_generate_tests(metafunc):
         return
 
     for param_name, values in profile.items():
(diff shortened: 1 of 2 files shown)


chaojun-zhang (approved):
LGTM

baodii (approved):
LGTM

Liangliang-Ma:
> will not run fused moe test, right?

Yes. Let's keep kernel level test now. Fused moe used some static tensor which works when model param unchanged.

PR #54, opened 2025-10-27
add op moe_align_block_size & batched_moe_align_block_size

mayuyuace:
Add op moe_align_block_size & batched_moe_align_block_size.
From: https://github.com/vllm-project/vllm/blob/main/csrc/moe/moe_align_sum_kernels.cu

### DIFF
--- a/csrc/moe/moe_align_sum_kernels.cpp
+++ b/csrc/moe/moe_align_sum_kernels.cpp
@@ -19,6 +19,7 @@ namespace vllm {
 namespace moe {
 
 constexpr int32_t WARP_SIZE = 32;
+constexpr int32_t EXCLUSIVE_SIZE = 1024;
 
 namespace batched_moe_align_block_size {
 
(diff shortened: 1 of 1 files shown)


jikunshang (approved):
LGTM!

csrc/moe/moe_align_sum_kernels.cpp
```diff
@@ -6,9 +6,291 @@
...
+
+  void operator()(sycl::nd_item<1> item) const {
+    auto group = item.get_group();
+    auto local_id_x = item.get_local_id(0);
+    auto local_range = item.get_local_range(0);
+
+    int32_t* temp_storage = static_cast<int32_t*>(
+        slm.template get_multi_ptr<sycl::access::decorated::no>().get());
+
+    int32_t* shared_counts = temp_storage + 1024;
```
jikunshang: why 1024 here? please avoid use magic number.

mayuyuace: Temp_storage needs 1024 int32 space.
CUDA does not need to display the declaration, but SYCL needs to display the declaration of this part of SLM.

mayuyuace: I see.
The const param named EXCLUSIVE_SIZE is added.

PR #58, opened 2025-11-04
[CHUNK_PREFILL] enable sink and local attn

YizhouZ:
* enable local attn, previous implementation is not fully functional.
* add softmax sink in kernel impl and template parameters.


jikunshang (approved):
LGTM.

PR #59, opened 2025-11-06
[FusedMoe] support fp16 grouped_gemm

mayuyuace:
Add data type template policy for moe_gemm, and support fp16.

### DIFF
--- a/csrc/xpu/cutlass_kernels/fused_moe_prologue.hpp
+++ b/csrc/xpu/cutlass_kernels/fused_moe_prologue.hpp
@@ -13,7 +13,8 @@ inline T ceilDiv(T a, T b) {
 
 // TODO: this function causes a build error
 // template <int RANGE_DIM, typename T>
-// inline void sycl_print_decimal(sycl::nd_item<RANGE_DIM> item, T* data, int size,
+// inline void sycl_print_decimal(sycl::nd_item<RANGE_DIM> item, T* data, int
+// size,
 //                                char* name) {
 //   int local_id = item.get_local_id(2);
 //   int group_id_x = item.get_group(2);


csrc/xpu/cutlass_kernels/grouped_gemm.hpp
```diff
@@ -8,15 +8,26 @@
...
 
 #include <torch/all.h>
+
+#include "collective/gemm/moe_dtype_policy.hpp"
 #include "utils.h"
 
 namespace gpu::cutlass_kernel {
 
 namespace grouped_gemm {
+template <class moe_policy>
```
Liangliang-Ma: We can do a template like this:
```
template<typename ElementType>
void kernel_functor

...

kernel_functor<cutlass::half_t>(args..)
```
ref:https://github.com/intel/sycl-tla/blob/main/examples/09_bmg_grouped_gemm_f8/09_bmg_grouped_gemm_f8.cpp#L559

csrc/xpu/cutlass_kernels/grouped_gemm_kernel.cpp
```diff
@@ -429,9 +425,13 @@ void kernel_functor(sycl::queue& stream, void* ptr_A, void* ptr_B,
   using GmemTiledCopyB =
       XE_2D_U16x32x32_LD_V;  // Note: This shape has to match the shape used for
                              // the scaling factors
+  using MMAOperation =
+      std::conditional_t<std::is_same_v<moe_policy, moe_bf16_policy>,
```
Liangliang-Ma: we better put this in policy.hpp too for afterwards other dtype support.

csrc/xpu/cutlass_kernels/grouped_gemm.hpp
Liangliang-Ma: You're right.

csrc/xpu/cutlass_kernels/grouped_gemm_kernel.cpp
mayuyuace: Done.

Liangliang-Ma (approved):

PR #69, opened 2025-11-12
support fp8 grouped gemm

mayuyuace:
Support wfp8a16.
And only support per tensor scale.

Now cutlass-sycl uses inefficient data type convert function from fp8 to other dtype:
https://github.com/intel/sycl-tla/blob/5ac97000432951b2812ade99235fbf539958c5d4/include/cutlass/float8.h#L336
image
As the picture shows, cuda will uses hardware instruction.
So the fp8 grouped gemm kernel has low performance.
Cutlass needs provide more efficient convert function.

**Update**:
Optimize dequantize fp8 by replacing native dtype convert function:
For example when m=1, n 5120, k=8192, e=16, topk=1 on **PVC**:
fp16 @ e5m2: 6.321ms -->  1.358ms
bf16 @ e5m2: 6.319ms -->  3.931ms
fp16 @ e4m3: 7.007ms -->  2.648ms
bf16 @ e4m3: 7.109ms -->  5.072ms

### DIFF
--- a/csrc/xpu/cutlass_kernels/collective/gemm/moe_array_mma.hpp
+++ b/csrc/xpu/cutlass_kernels/collective/gemm/moe_array_mma.hpp
@@ -55,6 +55,137 @@ struct MainloopMoE16Group {
 
 }  // namespace cutlass::gemm
 
+namespace cutlass {
+CUTLASS_HOST_DEVICE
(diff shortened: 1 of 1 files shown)


jikunshang:
I think we can on hold this now that we will upgrade sycl-tla v0.6 which may break current fp16/bf16 moe gemm..

PR #94, opened 2025-12-11
refactor grouped gemm and apply fused_moe fp8/int4/mxfp4

mayuyuace:
Unify moe_gemm and grouped_gemm into a single collective class.
And add UTs for fused_moe of bf16/fp16/fp8/int4/mxfp4.


Liangliang-Ma (changes requested):
We better extract your code to a mainloop/epilogue/scheduler collective, and wrap them in gemm_universal/gemm_universal_adapter, using the common frontend.

mayuyuace:
> You should extract your code to a mainloop/epilogue/scheduler collective, and wrap them in gemm_universal/gemm_universal_adapter, using the common frontend. Why you just rename some file and delete all previous moe code?

I wrote new collective classes, not only renamed files.

PR #113, opened 2026-01-05
[OneDNN] add vllm own dnnl stream and engine

zufangzhu:
Since PyTorch exposes onednn symbols through GpuStreamManager and GpuEngineManager instances, which create dnnl::engine and dnnl::stream during Torch initialization. To avoid mixing different ONEDNN libraries at runtime, create separate engines and streams for VLLM.


baodii (approved):
LGTM

zufangzhu:
> maybe we can add some ut to check that with ONEDNN_VERBOSE, the commit id is same to the one we are using? And we will still see 2 onednn version log when we enable torch op and vllm-xpu-kernel op at same time, right?

Yes, this aligns with our expectations. How can we know which version of ONEDNN is being used in Torch during each execution? Read from upstream github?

PR #115, opened 2026-01-05
[CHUNK_PREFILL] add seq_k support

YizhouZ:
when block_table is enabled, we must pass seq_k rather than cu_seq_k

### DIFF
--- a/.github/workflows/ut.yaml
+++ b/.github/workflows/ut.yaml
@@ -61,7 +61,7 @@ jobs:
   run-unit-tests-pvc:
     runs-on: self-hosted-pvc
     needs: build-docker-image-latest-pvc
-    timeout-minutes: 30
+    timeout-minutes: 40
     container:                      
       image: localhost:5000/xpu-kernel-ci-image:latest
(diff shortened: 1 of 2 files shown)


tests/flash_attn/test_flash_attn_varlen_func.py
```diff
@@ -174,9 +174,10 @@ def test_varlen_with_paged_kv(
     cu_query_lens = torch.tensor([0] + query_lens,
                                  dtype=torch.int32).cumsum(dim=0,
                                                            dtype=torch.int32)
-    cu_kv_lens = torch.tensor([0] + kv_lens,
-                              dtype=torch.int32).cumsum(dim=0,
-                                                        dtype=torch.int32)
+    cu_kv_lens = torch.tensor(  #noqa: F841
```
jikunshang: I feel we can remove this now?

YizhouZ: added ut to use this cu_kv_lens.

PR #117, opened 2026-01-07
add fused gdn kernel

mayuyuace:
bmg fp8 lm_eval results:
image

**Note**: Below performance with triton is selected from second running, because triton kernels need auto tunning when first running, which will spends more time. TTFT with triton path first running will be bigger than 10000ms on BMG.
Performance:
PVC fp16:
Throughtputs improved from **204** to **248**.
Triton path:
image
Replaced by this PR:
image
BMG fp8 (only test fp8 in BMG due to memory limit):
Throughtputs improved from **300** to **380**.
Triton path:
image
Replaced by this PR:
image

### DIFF
--- a/.github/workflows/ut.yaml
+++ b/.github/workflows/ut.yaml
@@ -95,7 +95,7 @@ jobs:
   run-unit-tests-bmg:
     runs-on: self-hosted-bmg
     needs: build-docker-image-latest-bmg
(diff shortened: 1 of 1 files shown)


mayuyuace:
Next work is to opitimize gated_delta_rule kernel to chunk kernel for prefill stage.

PR #123, opened 2026-01-14
add paged decode split kernels for VLLM 

baodii:
Add paged decode attention for VLLM. This PR is for pure decode phase. The kernel is specifically for long sequence. 
The performance for pure decode phase is as below:
image

### DIFF
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -94,7 +94,7 @@ std::vector<at::Tensor> mha_varlen_fwd(
   bool is_local = (window_size_left != -1) | (window_size_right != -1);
   bool is_sink = softmax_sink_.has_value();
 
-  if (max_seqlen_q > 1 || is_local || !is_paged || is_causal) {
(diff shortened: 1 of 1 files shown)


yma11:
verified using llama3.1-8B on single BMG card:
image
acc:
image

rogerxfeng8 (approved):
LGTM

csrc/xpu/attn/xe_2/kernel/paged_decode_kernel.hpp
```diff
@@ -0,0 +1,725 @@
+/***************************************************************************************************
+ * Copyright (C) 2025 Intel Corporation, All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
```
rogerxfeng8: Any reason to take BSD-3 license? Other files we are using Apache 2.0

baodii: Hi, Roger, I’ve identified multiple files that use this license. I suggest we align on the approach and address all required changes in a follow-up PR.

PR #124, opened 2026-01-15
[FusedMoE] input reorder kernel support low precision inputs with scales

Liangliang-Ma:
supports inputs with type bf16/fp16, mxfp8/mxfp4 with e8m0 scales, fp8 with 1x128 block scales.


baodii (approved):
LGTM

PR #133, opened 2026-01-30
MoE: Optimize and fix moe_align_block_size & moe_lora_align_block_size kernels

chaojun-zhang:
## Test Plan
pytest -s -v tests/test_moe_align_sum.py  
pytest -s -v tests/test_moe_sum.py 
pytest -s -v ests/test_moe_lora_align_sum.py
## Test Result
ALL passed

### DIFF
--- a/csrc/moe/moe_ops.h
+++ b/csrc/moe/moe_ops.h
@@ -61,7 +61,8 @@ void topk_softmax(
     torch::Tensor& topk_indices,
     torch::Tensor& token_expert_indices,
     torch::Tensor& gating_output,
-    const bool renormalize);
+    const bool renormalize,
+    std::optional<torch::Tensor> bias);
 
 void moe_gather(
     torch::Tensor& output,
--- a/csrc/moe/topk_softmax.cpp
+++ b/csrc/moe/topk_softmax.cpp
@@ -105,7 +105,8 @@ class moeTopK {
       const int k,
       const int start_expert,
       const int end_expert,
-      const bool renormalize)
+      const bool renormalize,
+      const float* bias)
       : inputs_after_softmax(inputs_after_softmax),
         finished(finished),
         output(output),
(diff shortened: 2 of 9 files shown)


csrc/moe/moe_align_sum_kernels.cpp
```diff
@@ -29,46 +30,46 @@ static constexpr int32_t num_blocks = 1;
...
+        num_tokens_post_pad(num_tokens_post_pad),
+        slm(slm) {}
 
   void operator()(sycl::nd_item<1> item) const {
-    // TODO: This is a naive implementation. Could be optimized.
-    auto group = item.get_group();
-    auto local_id_x = item.get_local_id(0);
-    auto local_range = item.get_local_range(0);
-    auto group_range = item.get_group_range(0);
+    auto threadIdx_x = item.get_local_id(0);
```
jikunshang: we'd better follow sycl, instead of using cuda thread/block/grid.

csrc/moe/moe_align_sum_kernels.cpp
```diff
@@ -125,145 +126,399 @@ class batched_moe_align_block_size_kernel {
...
+  // synchronization easier.
+  if (threadIdx_x < fill_threads) {
+    // Initialize sorted_token_ids with numel
+    for (size_t it = threadIdx_x; it < max_num_tokens_padded;
+         it += fill_threads) {
+      sorted_token_ids[sorted_token_ids_offset + it] =
+          static_cast<int32_t>(numel);
+    }
+    // Three __syncthreads() corresponding to the other threads
+    item.barrier(sycl::access::fence_space::local_space);
```
mayuyuace: Is three barrier really needed in sycl?

chaojun-zhang: updated, just use one barrier

PR #148, opened 2026-02-06
Support topk_sigmoid kernel for MoE

jerrychenhf:
Currently fused topk only support softmax function, this PR add topk sigmoid support.

MiniMax-M2 model uses topk sigmoid while there is only topk softmax kernel. It had to go with grouped topk with num_expert_group=1.  A recent change in vllm has fixed this (https://github.com/vllm-project/vllm/pull/31246) and therefore, we need the corresponding topk sigmoid kernel to run MiniMax-M2 on XPU. 

The changes:
- Support topk sigmoid kernel in MoE extension (the original topk_softmax.cpp to topk.cpp)
- Added unit test for topk sigmoid (the original test_topk_softmax.py to test_topk.py)
- Refactor the benchmark to support topk_sigmoid. (the original benchmark_topk_softmax.py to benchmark_topk.py)

### DIFF
--- a/.github/workflows/ut.yaml
+++ b/.github/workflows/ut.yaml
@@ -74,9 +74,8 @@ jobs:
       - name: build & install wheel
         run: |
           git config --global --add safe.directory "${GITHUB_WORKSPACE}"
-          python3 -m pip config set global.break-system-packages true
-          pip install -r requirements.txt
-          MAX_JOB=128 pip install --extra-index-url=https://download.pytorch.org/whl/xpu -e . -v
+          uv pip install -r requirements.txt
+          MAX_JOB=128 uv pip install -e . -v
 
       - name: test 
         run: |
@@ -109,9 +108,8 @@ jobs:
       - name: build & install wheel
         run: |
           git config --global --add safe.directory "${GITHUB_WORKSPACE}"
-          python3 -m pip config set global.break-system-packages true
-          pip install -r requirements.txt
-          MAX_JOBS=128 pip install --extra-index-url=https://download.pytorch.org/whl/xpu -e . -v
+          uv pip install -r requirements.txt
+          MAX_JOBS=128 uv pip install -e . -v
 
       - name: test 
         run: |
--- a/.github/workflows/wheel-per-commit.yaml
+++ b/.github/workflows/wheel-per-commit.yaml
@@ -42,8 +42,8 @@ jobs:
 
       - name: build wheel
         run: |
-          python3 -m pip config set global.break-system-packages true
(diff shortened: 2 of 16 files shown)


csrc/moe/topk.cpp
```diff
@@ -9,6 +9,18 @@ static constexpr int WARP_SIZE = 32;
 
 namespace vllm {
 namespace moe {
+enum class ScoringFunc {
+  SOFTMAX = 0,
+  SIGMOID = 1,
+};
+
+static inline float Sigmoid(float x) { return 1.0f / (1.0f + sycl::exp(-x)); }
```
jikunshang: function should use lower case.

csrc/moe/topk.cpp
```diff
@@ -748,16 +828,84 @@ void topk_softmax(
 
   const at::DeviceGuard device_guard(gating_output.device());
   auto& queue = vllm::xpu::vllmGetQueue();
-  torch::Tensor softmax_workspace = torch::empty(
+  torch::Tensor scoring_workspace = torch::empty(
+      {workspace_size}, gating_output.options().dtype(torch::kFloat));
+
+#define LAUNCH_TOPK(INPUTDTYPE, INDTYPE)                               \
```
jikunshang: why define this twice instead of making a unify macro?

csrc/moe/topk.cpp
```diff
@@ -9,6 +9,18 @@ static constexpr int WARP_SIZE = 32;
...
+  SOFTMAX = 0,
+  SIGMOID = 1,
+};
+
+static inline float Sigmoid(float x) { return 1.0f / (1.0f + sycl::exp(-x)); }
+
+template <typename T>
+static inline float SigmoidTyped(T x) {
+  const float val = Sigmoid(static_cast<float>(x));
+  return static_cast<float>(static_cast<T>(val));
```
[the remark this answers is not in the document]
jerrychenhf: The is for the case that T is not float type (BF16 or Float16). We need the accuracy to be consistent to the origin data type. The cast to T and back to float make sure of this. The topk results will be different if the data type accuracy doesn't match. This is why there are some test cases failed for BF16 and Float16 but float type cases are all passed.

csrc/moe/topk.cpp
jerrychenhf: There are pros and cons for both ways. For not using the same LAUNCH_TOPK macro, we actually reuse and save to passing every LAUNCH_TOPK  with SCROING function. (It's another way of reuse).
For using the same LAUNCH_TOPK, we are passing SCROING function for every call of LAUNCH_TOPK.

jikunshang:
@codex review

PR #166, opened 2026-03-01
Support FP8 KV cache in paged_decode kernel

baodii:
Add FP8 (e4m3/e5m2) KV cache support to the paged_decode kernel

- Add k_scale/v_scale to paged_decode_args_t
- Change dispatch from CutlassDType to CutlassQKType for mixed Q/KV types
- Add FP8 element type branches in decode_policy_dispatch_impl
- Add scale_k/scale_v to DecodeFwdMainloop::Arguments
- Add Fp8KV detection and on-the-fly dequantization in decode mainloop
- Pass k_scale/v_scale through the full call chain
- Remove is_fp8kv routing constraint in flash_api.cpp so paged_decode handles FP8 KV directly instead of falling back to chunked_prefill

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,6 +152,22 @@ if(VLLM_GPU_LANG STREQUAL "SYCL")
   set(SYCL_FIRST_HEADER "${CMAKE_CURRENT_SOURCE_DIR}/csrc/sycl_first.h")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -include ${SYCL_FIRST_HEADER}")
 
+  # AOT devices can be overridden via environment variables:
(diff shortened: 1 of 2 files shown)


csrc/xpu/attn/xe_2/collective/chunk_prefill_mainloop.hpp
```diff
@@ -664,10 +667,14 @@ struct DecodeFwdMainloop<
...
   static constexpr bool PagedKV = PagedKV_;
   static constexpr bool CausalMask = CausalMask_;
+  static constexpr bool Fp8KV =
+      is_any_of_v<ElementK, float_e5m2_t, float_e4m3_t>;
 
   // User-facing arguments
   struct Arguments {
     ElementS const scale;
+    ElementS const scale_k;
+    ElementS const scale_v;
```
xinyu-intel: scale_k and scale_v should be device ptr instead of host scalar.

PR #171, opened 2026-03-04
Add block_size 16/32 support for chunk prefill and fix paged decode

baodii:
Add page-size-aware chunk prefill policies with smaller K-tiles to support block_size 16 and 32 without division-by-zero in get_paged_idx() where tiles_per_page = page_size / K_tile would yield 0 when K_tile > page_size:

- fmha_utils.hpp: Add 7 new chunk prefill policy structs (5 for p16 with K_tile=16, 2 for p32 with K_tile=32 for head96/128)
- chunk_prefill_extern.hpp: Add paged-only extern template declarations
- chunk_prefill_configure.cmake: Add paged-only CMake generation loop
- fmha_xe2.cpp: Route to p16/p32 policies based on block_size when paged

Fix paged decode kernel issues for small block sizes:

- flash_api.cpp: Cap num_kv_splits to block_size to prevent overflow of fixed-size arrays in ReduceSplitK (max_num_kv_splits = SGPerWG * sg_size = block_size for all decode policies)
- chunk_prefill_epilogue.hpp: Initialize rA_max from tA_max in DecodeFwdEpilogue::reduce_A when ReduceK==1 (block_size=16) so that max_logits contains correct values for ReduceSplitK reduction


xinyu-intel (approved):

YizhouZ (approved):

rogerxfeng8:
What's the kernel binary size increase after the support of 26/32 block size?

jikunshang:
please close this if no longer needed.

baodii:
Implemented in #308

PR #186, opened 2026-03-09
[Kernel] use different options for diff kernels

zufangzhu:
1. remove AOT option `-internal_options -cl-intel-256-GRF-per-thread` for sycl kernel.
    kernel  like  fused_rms_norm, rotary_embedding and activation gain perf above 40%
2. refine AOT option, currently support
a. default, will build `pvc,bmg,bmg-g21-a0,bmg-g31-a0`
b. use env var set manually by: `VLLM_XPU_AOT_DEVICES="bmg" VLLM_XPU_XE2_AOT_DEVICES="bmg"`
c. not use aot: `VLLM_XPU_AOT_DEVICES="" VLLM_XPU_XE2_AOT_DEVICES=""`

- use different compile option for element-wise and attn kernels
image


baodii (approved):
LGTM

jikunshang (approved):

PR #204, opened 2026-03-19
[Decode Attn] Change strategy of num_splits to avoid acc issue

baodii:
### Problem

The split-KV decode attention path had two precision issues:
1. **Divide-multiply roundtrip**: The epilogue divided partial outputs by `exp_sum` (quantizing to bf16), then ReduceSplitK multiplied back by `exp_sum`. This roundtrip amplified bf16 quantization error, especially for outputs with large magnitudes.
2. **Mixed-batch short sequences**: When `num_kv_splits > 1`, even short-context sequences in a batch were split, incurring unnecessary precision loss from the reduce roundtrip.

Additionally, the `get_num_splits()` heuristic needed tuning to work well across different GPU architectures (BMG with 20 Xe-cores) and block sizes (64/128).

### Changes

**Precision improvements** (`chunk_prefill_epilogue.hpp`, `paged_decode_kernel.hpp`):
- **Unnormalized partial outputs**: When `num_kv_splits > 1`, the epilogue now stores raw (unnormalized) accumulator values to `Oaccum`. ReduceSplitK combines them directly without the multiply-back step, eliminating the bf16 divide-multiply amplification.
- **Per-sequence split decisions**: Short sequences (`< 128` KV blocks) in a batch are treated as single-split even when `num_kv_splits > 1`. Split 0 processes all blocks; splits 1+ skip entirely. The epilogue writes sentinel values (`exp_sums=1`, `max_logits=0`) so ReduceSplitK acts as a pass-through copy.
- **Safe buffer initialization** (`flash_api.cpp`): `max_logits` initialized to `-inf` and `exp_sums` to `0`, ensuring unused split slots are safely skipped by the `exp_sum ≤ 0` guard in ReduceSplitK.

**Heuristic tuning** (`flash_api.cpp`):
- Three-zone design based on `parallel / xe_cores` ratio:
  - **Under-utilized** (`par < xe_cores`): Split to fill GPU cores, with block_size-aware scaling.
  - **Well-utilized** (`par ∈ [xe_cores, 2×xe_cores]`): Don't split — GPU is busy, splitting adds overhead.
  - **Heavily oversubscribed** (`par > 2×xe_cores`): Split to shorten WGs, capped to avoid excessive scheduling overhead.
- Hard cap of 8 splits max; minimum 32 KV blocks per split; minimum 128 KV blocks to enable splitting.
- `xe_cores` queried at runtime from device, so the heuristic adapts to both PVC and BMG.

### DIFF
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -31,6 +31,7 @@ inline int get_num_splits(
     // Under-utilized: fill GPU cores.
     // Scale by block_size since larger blocks mean more compute per WG.
     int eff_parallel = cur_parallel * block_size / 64;
+    eff_parallel = std::max(1, eff_parallel);
     target_splits = (num_xe_cores + eff_parallel - 1) / eff_parallel;
   } else if (cur_parallel <= num_xe_cores * 2) {
     // Well-utilized zone (1x-2x oversubscription):


baodii:
> Thanks for fixing. QQ: G31(which should have 32 xe-core) whether affected?

It should be fine.

PR #228, opened 2026-03-27
 support topk topp sampler

mayuyuace:
Topk Topp sampler sycl kernels to fuse whole process of forward_native.
Add op exponential_2d_ to replace exponential_ of torch.
And this PR only support one seed for whole input, next PR will support different seeds for different batch.
Add benchmark for sycl topk topp: benchmarks/benchmark_topk_topp.py

BMG benchmark:
VS triton:
image
VS pytorch:
image


csrc/xpu/sampler/topk_topp_sampler_kernels.hpp
```diff
@@ -0,0 +1,1389 @@
...
+    const int local_id = item.get_local_linear_id();
+    const int local_range = item.get_local_range(0);
+
+    const int global_id = item.get_global_linear_id();
+    uint64_t philox_seed = seed;
+    uint64_t philox_offset = offset;
+    randStatePhilox4_32_10_t state;
+    rand_init(philox_seed, global_id, philox_offset, &state);
+
+    Uniform4DistributionFunctor dist_func;
```
jikunshang: minor: add namespace `RAND::Uniform4DistributionFunctor`

yma11:
> merge this to unblock vllm side change/test. just like attention kernel is keeping improvement.

I am confused. As you said, lm_eval won't use this path so can't test this kernel. Can you point out what kind of test can be done after merge? will it make more sense we use this method to have a small scope test first?

jikunshang (approved):
please revert file mode.
Overall LGTM, thanks!

xinyu-intel (approved):

PR #240, opened 2026-03-31
WA: route XE3/XE3P platforms to XE2 cutlass kernels

baodii:
Add is_xe3_arch() helper to detect Panther Lake (ptl_h, ptl_u) and Wildcat Lake (wcl) GPUs. Route these XE3/XE3P architectures to the existing XE2 cutlass kernel paths for paged decode so they can run instead of hitting TORCH_CHECK.


jikunshang (approved):
LGTM. 
minor: we could add LNL for Xe2 as well.

YizhouZ (approved):

KyleSanderson:
> LGTM. minor: we could add LNL for Xe2 as well.

ha - just hit this (on LNL), and confirmed fixed on v0.15.0.

Thanks all.

PR #256, opened 2026-04-08
skip scales check

mayuyuace:
G31 CI may raise random error, add equal_nan and print log.
The error cannot be reproduced on local machine.
Skip scales check for now.


jikunshang (approved):
Thanks for fixing.

xinyu-intel (approved):

PR #262, opened 2026-04-09
Fix use-after-free in get_xpu_view_from_cpu_tensor

chaojun-zhang:

## Purpose
**Problem**
  get_xpu_view_from_cpu_tensor creates a zero-copy XPU view over a CPU pinned-memory tensor by storing the raw host_ptr inside XPUHostViewAllocator. However, it held no reference
  to the original cpu_tensor, so its reference count could drop to zero and its pinned memory could be freed while the XPU view was still alive — a classic use-after-free.

  This is not a theoretical concern. In two real call sites inside vLLM:

   - [vllm/model_executor/offloader/uva.py](https://github.com/vllm-project/vllm/blob/6c749399b7dc939a592e30d5964d7bdf255427aa/vllm/model_executor/offloader/uva.py#L100) — cpu_data is a local variable created via .pin_memory(); after the enclosing function returns, Python GC frees it while p.data (the XPU
  view) still points to its memory.
   - [vllm/model_executor/model_loader/utils.py](https://github.com/vllm-project/vllm/blob/6c749399b7dc939a592e30d5964d7bdf255427aa/vllm/model_executor/model_loader/utils.py#L164) — same pattern; cpu_data is a local variable with no other owners.

  **Fix**

  XPUHostViewAllocator now holds a torch::Tensor owner_ member that keeps the source tensor's reference count elevated. The owner is stored as an OwnerContext in the DataPtr, so
  it lives exactly as long as the XPU view tensor does and is released when the view is destroyed.

## Test Plan
pytest -s -v tests/test_uva.py::test_view_lifetime_after_owner_drop

### DIFF
--- a/csrc/xpu_view.cpp
+++ b/csrc/xpu_view.cpp
@@ -2,6 +2,7 @@
 #include "ops.h"
 #include <c10/core/Device.h>
 #include <c10/xpu/XPUFunctions.h>
+#include <memory>
 
 namespace vllm::xpu {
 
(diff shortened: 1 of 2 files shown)


hsubramony (approved):

Liangliang-Ma (approved):

baodii (approved):
LGTM

PR #269, opened 2026-04-13
Fix overflow of remap_hidden_states when rows is huge

mayuyuace:
Replace int of ptr accumulation by int64 to avoid overflow.


tests/fused_moe/test_remap_hidden_states.py
```diff
@@ -226,6 +226,98 @@ def test_remap_hidden_states(num_rows, hidden_size, total_experts_num, topk,
             print("Mismatched ref:", ref_unpermuted_scales[mismatched_indices])
```
jikunshang, line 226: can we revert this part now?

jikunshang (approved):
LGTM. please rename PR title. thanks!

xinyu-intel (approved):

tests/fused_moe/test_remap_hidden_states.py
mayuyuace: This PR is for https://jira.devtools.intel.com/browse/VLLMZ-982 .
Not related with the scale part.

PR #285, opened 2026-04-17
Fix moe_align kernel bug

mayuyuace:
Sometimes the comparation of int and size_t returns wrong result.
Fix the kernel by align dtype.


xinyu-intel (approved):

xinyu-intel:
consider to clean up the int in the whole repo. size_t is good to cover combinations of compiler and os though we stick to linux and gcc. @jikunshang

jikunshang:
> consider to clean up the int in the whole repo. size_t is good to cover combinations of compiler and os though we stick to linux and gcc. @jikunshang

make sense. let's ask agent do that.:)

jikunshang (approved):

PR #292, opened 2026-04-20
[Cache] support HND in reshape_and_cache_flash

zufangzhu:
reshape_and_cache_flash_strided_kernel support HND, only add test.


jikunshang:
can we close this

PR #295, opened 2026-04-20
[Refactor] refactor vectorization

zufangzhu:
- use dynamic (larger) vectorization in reshape_and_cache_flash kernel

For KPI 4 models shapes, after refactor, this kernel almost all have better performance than before, especially in the decode phase:
image

### DIFF
--- a/csrc/cache.cpp
+++ b/csrc/cache.cpp
@@ -6,6 +6,7 @@
 
 #include "dispatch_utils.h"
 #include "quantization/fp8/quant_utils.h"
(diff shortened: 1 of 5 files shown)


xinyu-intel (approved):

jikunshang (approved):

zufangzhu:
Thanks dhiraj.d.kalamkar@intel.com and sundaresan.g@intel.com for the help to provide ideas of optimizing this kernel.

PR #296, opened 2026-04-20
Add moe dynamic quant kernels

tvoas:
This kernel adds optimized ops for moe_scatter_dynamic_quant and moe_swiglu_dynamic_quant, following the request by ByteDance for support for these ops. In performance measurements, these ops were within a few percent of the same kernels on IPEX, and in most cases exceeded IPEX's bandwidth.

### DIFF
--- a/csrc/moe/dynamic_quant_kernels.cpp
+++ b/csrc/moe/dynamic_quant_kernels.cpp
@@ -11,6 +11,139 @@
 using namespace sycl::ext::intel::esimd;
 using bf16 = sycl::ext::oneapi::bfloat16;
 
+void moe_swiglu_dynamic_quant(
+    torch::Tensor& scatter_tokens,
+    torch::Tensor& smooth_scale,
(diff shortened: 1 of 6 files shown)


xinyu-intel (changes requested):
I don't think we should support int8 quantization, can you pls try with fp8/mxfp8/mxfp4?

tvoas:
> I don't think we should support int8 quantization, can you pls try with fp8/mxfp8/mxfp4?

Right now support for bfloat16 to int8 was added following the [requested benchmark support from ByteDance](https://github.com/abenmao/xpu-perf/blob/fdb928cbdf40ab1556dc5a0c9efb40b76b2e7606/micro_perf/workloads/llm/igpu_ops/moe_ops.json#L33). Its also what IPEX supported. I can try adding additional datatype support. Supporting float16 source data should be simple. And supporting fp8 destination data I also think is doable though needs a bit more work. The MX formats, and especially mxfp4 would require significant changes.

PR #308, opened 2026-04-23
[decode][attn] Enable 16/32/64xn block size for attention

baodii:
## Summary

The `kv_tile=_128` decode policy uses `SubgroupLayoutQK<_1, _8, _1>`
(8 subgroups reducing across K, `ReduceK = 8`), which exercises a
buggy cross-SG SLM reduction path in
`chunk_prefill_epilogue.hpp::reduce_A`. The bug only manifests under
real autoregressive vLLM serving — standalone unit tests with the same
shapes/strides pass — and stems from the degenerate
`SGTileShapeO = (1, 32)` / `ReduceFragA = (32, 1)` layout that results
when `ReduceK = 8` splits the `(Q = 8, V = 32)` per-WG output among
8 subgroups.

This PR routes any `page_size` that is a multiple of 64 (including
multiples of 128) through the `kv_tile = _64` policy, which uses
`ReduceK = 4` and the well-tested `SGTileShapeO = (2, 32)` layout. The
mainloop already iterates `page_size / 64` sub-tiles per page via the
page-table indirection (2 for `page = 128`, 4 for `page = 256`, ...),
so correctness is preserved at a small parallelism cost. The `_128`
specialization is kept (marked unused) so it can be re-enabled once
the upstream `ReduceK = 8` reduction path is fixed.

## Impact (gpt-oss-20b, gsm8k, Intel Arc Pro B60)

| Configuration                  | strict-match | flexible-extract |
| ------------------------------ | ------------ | ---------------- |
| `block_size = 64`              | 82.25 %      | 94.50 %          |
| `block_size = 128` (before PR) | 68.00 %      | 80.00 %          |
| `block_size = 128` (this PR)   | 79.75 %      | 92.25 %          |

The ~12 pp strict-match drop with `block_size = 128` was a hard blocker
for production serving with larger block sizes; this PR restores
parity with `block_size = 64`.

## Dispatch table after this PR

| `block_size` | dispatched policy | sub-tiles per page |
| ------------ | ----------------- | ------------------ |
| 16           | `_16`             | 1                  |
| 32           | `_32`             | 1                  |
| 64           | `_64`             | 1                  |
| 128          | `_64`             | 2                  |
| 192          | `_64`             | 3                  |
| 256          | `_64`             | 4                  |
| 320, 384, …  | `_64`             | 5, 6, …            |

## Test changes

The `head_size == 512 and block_size >= 128` skip in
`test_decode_with_paged_kv` is no longer warranted: SLM usage is now
independent of `block_size` for `block_size >= 64`. Verified that
`head_size = 512` passes for all `BLOCK_SIZES` (16, 32, 64, 128, 192,
256) across all `is_sink` / `is_casual` / sliding-window variants in
the decode test (96 cases total — all pass).

## Files changed

- `csrc/xpu/attn/xe_2/paged_decode_utils.hpp` — collapse the
  `page_size % 128` branch into the `page_size % 64` branch in
  `dispatch_by_page_size`.
- `csrc/xpu/attn/xe_2/fmha_utils.hpp` — mark the `kv_tile=_128`
  decode policy specialization as currently unused with a pointer to
  the underlying bug; update the `_64` policy comment to reflect that
  it now services all 64-multiples.
- `tests/flash_attn/test_flash_attn_varlen_func.py` — drop the
  obsolete `head_size == 512 and block_size >= 128` skip in
  `test_decode_with_paged_kv`.

## Follow-up

The actual `ReduceK = 8` SLM-reduction bug should be fixed upstream
so the `_128` decode policy can be re-enabled for higher decode
throughput at large block sizes. A type-level diff and suspected
primitives are noted in the in-tree comments.


jikunshang:
please fix conflict, thanks.

jikunshang (approved):

csrc/xpu/attn/xe_2/fmha_utils.hpp
YizhouZ: we need to change this head96 to block_size=32 to avoid block32head96 cases not supported.

YizhouZ (approved):

jikunshang:
wheel size is 280MB now. we need consider how to improve. let's merge this once CI passed.

tests/flash_attn/test_flash_attn_varlen_func.py
```diff
@@ -11,12 +11,10 @@
 from vllm_xpu_kernels.flash_attn_interface import flash_attn_varlen_func
 
 NUM_HEADS = [(8, 2)]
-HEAD_SIZES = [64, 128, 192, 256, 512]
-BLOCK_SIZES = [64, 128]
+HEAD_SIZES = [64, 128, 256, 512]
+BLOCK_SIZES = [16, 64, 512]
```
jikunshang, line 15: should we add block size 512?

PR #312, opened 2026-04-24
Fix workgroup barrier deadlock

frost-intel:
## Purpose
Every subgroup in the workgroup should execute the same number of K-loops. The Xe2 FMHA kernel computes `k_block0`, `k_blocks`, and `k_blocks_causal` from `seq_coord`, which depends on `q_offset_sg` — a per-subgroup broadcast of the thread's row coordinate. Under causal masking, different subgroups within the same WG compute different loop bounds and execute different iteration counts, leaving some subgroups stuck at `barrier_wait` forever.

As a concrete example:
`(seq_q=129, seq_k=463)` causal with `TileQ=128`, `tile_k=64`:

SG0 (row 0) -> `k_blocks = 6`
SG7 (row 112) -> `k_blocks = 8`

**I'm fairly confident that this is the root cause of the hang in PVC flash_attn kernels. I'm not sure why this hasn't been an issue on BMG**. However I'm not an expert so I'd welcome any feedback on this solution.

This fix computes tight per-SG bounds as before, then reduces across the workgroup to obtain uniform bounds for the loop. This change resolves the hang in the xe_2 FMHA chunk-prefill kernel that occurred under causal masking with short variable-length q, and under sliding-window (local) masking. These are the cases currently guarded by SKIP_HANG_KERNEL in tests/flash_attn/test_flash_attn_varlen_func.py.

## Test Plan
`pytest -v -s tests/flash_attn/`
Note this no longer requires SKIP_HANG_KERNEL=1

## Test Result
All tests pass, no hang.

### DIFF
--- a/csrc/xpu/attn/xe_2/kernel/chunk_prefill_kernel.hpp
+++ b/csrc/xpu/attn/xe_2/kernel/chunk_prefill_kernel.hpp
@@ -272,7 +272,7 @@ class XeFMHAFwdKernel {
           cute::min(seq_len_qo, (blk_q * get<0>(TileShapeQK{}) + q_offset_sg));
 
       // calc sg level seq_len_kv
-      const int sg_seq_len =
+      const int seq_len =
           CausalMask
               ? LocalMask
                     ? cute::min(
@@ -289,12 +289,10 @@ class XeFMHAFwdKernel {
                     0) /
                     get<1>(TileShapeQK{})
               : 0;
-      const int sg_k_blocks =
-          cute::ceil_div(sg_seq_len, get<1>(TileShapeQK{}));
+      const int sg_k_blocks = cute::ceil_div(seq_len, get<1>(TileShapeQK{}));
       const int sg_k_blocks_causal =
-          CausalMask
-              ? (seq_coord + full_tile_offset) / get<1>(TileShapeQK{})
-              : 0;
+          CausalMask ? (seq_coord + full_tile_offset) / get<1>(TileShapeQK{})
+                     : 0;
 
       // The mainloop wraps each K iteration in a workgroup-scoped barrier
       // pair, so every subgroup in the workgroup must execute the same
@@ -376,7 +374,7 @@ class XeFMHAFwdKernel {
           k_blocks,
           k_blocks_causal,
           thr_id,
-          sg_seq_len,
+          seq_len,
           full_tile_offset);
 
       // return softmax_lse
--- a/csrc/xpu/attn/xe_2/kernel/chunk_prefill_kernel.hpp
+++ b/csrc/xpu/attn/xe_2/kernel/chunk_prefill_kernel.hpp
@@ -272,7 +272,7 @@ class XeFMHAFwdKernel {
           cute::min(seq_len_qo, (blk_q * get<0>(TileShapeQK{}) + q_offset_sg));
 
       // calc sg level seq_len_kv
-      const int seq_len =
+      const int sg_seq_len =
           CausalMask
               ? LocalMask
                     ? cute::min(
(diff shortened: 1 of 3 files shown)


frost-intel:
@YizhouZ Can you review this?

jikunshang:
cc @xuechendi

csrc/xpu/attn/xe_2/kernel/chunk_prefill_kernel.hpp
```diff
@@ -275,18 +275,38 @@ class XeFMHAFwdKernel {
...
+      // K-loop trip count. Reduce the per-SG bounds across the WG:
+      //   k_block0        = min across WG (start no later than any SG)
+      //   k_blocks        = max across WG (end no earlier than any SG)
+      //   k_blocks_causal = min across WG (turn on causal masking no later
+      //                                    than any SG needs it)
+      // Per-element causal / local / remainder masking inside the mainloop
+      // handles the widened range safely for SGs that didn't need it.
+      auto wg = sycl::ext::oneapi::this_work_item::get_work_group<3>();
+      const int k_block0 = LocalMask
+          ? sycl::reduce_over_group(wg, sg_k_block0, sycl::minimum<int>{})
```
baodii: @YizhouZ please figure out how much will this reduce op impact on perf.

frost-intel:
I ran the `benchmark_cutlass_flash_attn_varlen.py` tests on B60, and I found that the causal MHA with (16, 16) heads had an 8-10% perf regression from K-loop widening.

The reduce_over_group(max) on k_blocks forces all 16 subgroups in the workgroup to iterate the K range needed by the bottom-most SG (the one with the most Q rows, hence the longest causal K range). On boundary tiles, the topmost SG can end up executing ~7 extra K blocks of full GEMM1 + GEMM2 that get entirely masked out by causal masking.

My fix is to pass the per-SG bounds (sg_k_block0, sg_k_blocks) into the mainloop alongside the unified bounds. Guard the GEMM body with if (K >= sg_blk_k0 && K < sg_blk_k1) while keeping both barriers unconditional. SGs outside their range skip compute/loads but still participate in synchronization.

This completely eliminated any perf difference from this PR. The accuracy tests all had identical results to the main branch, and timing results are within +-2% of all the timing from the main branch, and up to 6% for some of the causal MHA configs (such as 16, 16 heads).

frost-intel:
I ran `benchmark_cutlass_flash_attn_varlen.py` (228 configs) on a B60 (112 configs — the subset where a regression was originally reported). **6 runs on `main`, 4 runs on this PR**, then compared with paired t-tests.

#### Results

Mean end-to-end latency across all 112 configs:

| Branch | Mean latency |
|---|---|
| `main` (×6 runs) | 10,361 – 10,370 µs |
| This PR (×4 runs) | 10,367 – 10,382 µs |

The difference is **< 0.2%** and driven by two configs that exhibit bimodal latency (~4,360 µs vs ~4,760–4,865 µs) due to a hardware/thermal transient. Critically, `main` also hits the slow mode for both configs (e.g. main run 5 at 4,775 µs). Excluding those two configs, all remaining 110 configs vary by ≤ 2% with no systematic per-branch pattern.

The one pairwise comparison that crossed the t-test significance threshold (p=0.016, main[0] vs PR[6]) is entirely explained by PR[6] landing in the slow hardware mode for those two configs.

#### Conclusion

There is **no regression**. The variation is hardware noise: two configs have a bimodal performance state that affects both branches equally. With more repeated runs the effect washes out — the PR mean is within 0.2% of main across all 10 runs.

Additionally, accuracy results were identical before/after this PR.

frost-intel:
@jikunshang @YizhouZ @baodii Could this be added to 0.19 release?

jikunshang:
please resolve DCO issue and pre commit issue first, thanks

YizhouZ (approved):

baodii (approved):

xuechendi:
@frost-intel , thanks, I can confirm the hung issue resolved on PVC with flash_attn.

PR #313, opened 2026-04-27
Reduce ops of moe and optimize RowsPerExpertCount

mayuyuace:
To reduce CPU time when bs is small.
1. remove copy of topk ids
2. fuse CalculateFristTokenOffset to RemapHiddenStates
3. in fused_moe, use rows_per_expert to replace expert_first_token_offset.

Kernel Optimization: 
For RowsPerExpertCount kernels, by using hierarchical reduction, kernel perf gets much improvement. Like M=8192,  NUM_EXPERTS = 128, TOP_K = 4, device time reduced from 138.437us to 11.66 us.


Liangliang-Ma (approved):

zufangzhu (commented):
Can you provide a profiling result of a moe model in small bs

mayuyuace:
@zufangzhu 
GPT-OSS 120b with 16 batch size.
image

zufangzhu (approved):
TPOT 0.0354 -> 0.0338 
LGTM!

PR #315, opened 2026-04-28
Fix rope kernel OOB issue

Yejing-Lai:
Fix  zai-org/glm-4-9b, zai-org/glm-4v-9b and 01-ai/Yi1.5-34B-Chat lost device with input/output:8K/2K
JIRA track: https://jira.devtools.intel.com/browse/VLLMZ-1145

Root cause: These models do not support rope_scaling.  It results in out-of-bounds read within the RoPE kernel when online-serving with max_model_len > model_seq_length.


Yejing-Lai:
To align with the CUDA RoPE kernel code, this issue will be fixed on the vLLM side.
https://github.com/intel-innersource/applications.ai.gpu.vllm-xpu/pull/179

PR #317, opened 2026-04-29
fix the bias missing issue when run fp8/fp4 gemm benchmark

czhu15:
This PR fixed below error:
"TypeError: fp8_gemm() missing 1 required positional argument: 'bias' "
when run below test case:
"python benchmark_gemm_onednn.py --benchmarks fp8"

Similar issue for fp4 was fixed too.

### DIFF
--- a/tests/test_fp4_gemm_onednn.py
+++ b/tests/test_fp4_gemm_onednn.py
@@ -68,7 +68,6 @@ def test_mxfp4_gemm(mnk_factors, out_dtype):
         inputs_scale,
         weights_scale,
         out_dtype,
-        torch.Tensor(),
     )
 
     output_ref = torch.matmul(inputs_hp.to(out_dtype),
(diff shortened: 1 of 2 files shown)


tests/test_fp4_gemm_onednn.py
```diff
@@ -68,7 +68,6 @@ def test_mxfp4_gemm(mnk_factors, out_dtype):
         inputs_scale,
         weights_scale,
         out_dtype,
-        torch.Tensor(),
```
jikunshang, line 71: passing an empty tensor here works? I think we should pass a None instead?

czhu15: the old code used an empty tensor, which works but IMO is a bit confusing. I prefer to pass a None instead, which is the default value if user does not input this argument.

xinyu-intel (approved):

jikunshang (approved):

Issue #319, opened 2026-04-29
Tests incompatible with Python < 3.12


mzweilin:
https://github.com/vllm-project/vllm-xpu-kernels/blob/7048b1f9464985fcae269a95e653713fe85a7f48/tests/test_indexer_k_quant_and_cache.py#L181-L182

The SyntaxError: f-string expression part cannot include a backslash occurs in Python versions 3.11 and earlier because backslashes were strictly forbidden inside the curly braces {} of an f-string. This restriction was lifted in Python 3.12. 

However, README states that the project works on Python 3.9 - 3.12

https://github.com/vllm-project/vllm-xpu-kernels/blob/7048b1f9464985fcae269a95e653713fe85a7f48/README.md?plain=1#L44

jikunshang:
@LLee233 can you take a look?

jikunshang:
see #323

PR #323, opened 2026-05-05
Align expression for broader python version

LLee233:
## Purpose
Align expression for broader python versions. For fixing #319

### DIFF
--- a/tests/test_indexer_k_quant_and_cache.py
+++ b/tests/test_indexer_k_quant_and_cache.py
@@ -175,11 +175,11 @@ def test_indexer_k_quant_and_cache(num_tokens, head_dim, quant_block_size,
         ref_scale = block_ref[fp8_end:].view(torch.float32)
         out_scale = block_xpu[fp8_end:].view(torch.float32)
 
+        diff = (ref_fp8.view(torch.float8_e4m3fn).float() -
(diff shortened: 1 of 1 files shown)


jikunshang (approved):

tests/test_indexer_k_quant_and_cache.py
```diff
@@ -175,11 +175,11 @@ def test_indexer_k_quant_and_cache(num_tokens, head_dim, quant_block_size,
         ref_scale = block_ref[fp8_end:].view(torch.float32)
         out_scale = block_xpu[fp8_end:].view(torch.float32)
 
+        diff = (ref_fp8.view(torch.float8_e4m3fn).float() -
+                out_fp8.view(torch.float8_e4m3fn).float()).abs().max()
```
xinyu-intel, line 179: why not directly upcast to float here?

LLee233: Do you mean the ‘.to()’ part? Because the results were stored in form of uint8 raw bytes. It should be turned back to fp8 and then do the comparing thing. Upcast to fp32 may change the original data I think.

xinyu-intel: thx.

xinyu-intel (approved):

PR #324, opened 2026-05-06
refactor template gen

jikunshang:
## Purpose

This PR introduces a configurable kernel compilation system for attention kernels (chunk prefill and paged decode), significantly reducing build time and binary size by allowing users to compile only the kernel variants needed for their models. It also improves the developer experience with better error messages, a Python fallback path, 

### Changes
1. Kernel Configuration Template System
Introduces VLLM_PAGED_DECODE_CONFIG and VLLM_CHUNK_PREFILL_CONFIG CMake options (also forwarded via [setup.py](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) from environment variables) that select which kernel variants to compile.

- New config files in kernel_configs
  - paged_decode_default.conf / paged_decode_full.conf
  - chunk_prefill_default.conf / chunk_prefill_full.conf
- CMake-generated headers (chunk_prefill_enabled_policies_gen.hpp, paged_decode_enabled_policies_gen.hpp) encode the enabled policy set at build time
- Template files (.hpp.in) drive the code generation via configure_file()
- Resolves shorthand names (paged_decode_llama) to full paths automatically

Usage 
```
# default: still full build
pip install --no-build-isolation .

# use default config, with a minimal build scope:
VLLM_CHUNK_PREFILL_CONFIG=chunk_prefill_default.conf \
VLLM_PAGED_DECODE_CONFIG=paged_decode_default.conf \
  pip install --no-build-isolation  .

```
2. Actionable Runtime Error Messages
When a kernel variant is not compiled, the runtime error now tells the user exactly which config line to add and how to rebuild, instead of a cryptic template error:
```
Chunk prefill kernel tuple not compiled for this configuration.

Add this line to your chunk_prefill config file:
  128,true,true,false,false,false

Then rebuild:
  VLLM_CHUNK_PREFILL_CONFIG=chunk_prefill_default.conf pip install .
```

3. Python Fallback for Missing Kernels (flash_attn_interface.py)
- Added ref_paged_attn(): a pure-PyTorch reference attention implementation supporting paged/non-paged, causal, local window, soft-cap, FP8 KV, and softmax LSE output
- flash_attn_varlen_func() now catches RuntimeError("not compiled ...") and falls back to ref_paged_attn() with a warning, preventing hard crashes during development

Limitations
- Default config is minimal: paged_decode_default.conf and chunk_prefill_default.conf cover only a small set of common configurations (Llama + DeepSeek MLA). Models with unusual head sizes, page sizes, or attention modes may hit a "not compiled" error and need a custom config or the full preset.
- Python fallback is for development only: ref_paged_attn() is a pure PyTorch reference; performance is orders of magnitude slower than the XPU kernel. It is not suitable for production inference.
- Kernel template system is XE2-only: The configurable compilation currently applies only to XE2 attention kernels . Other kernel types are not yet configurable.


csrc/xpu/attn/kernel_configs/paged_decode_default.conf
xinyu-intel, line 1: can config files be moved to a single place like cmake/config/?

jikunshang: this should be shared across attention kernel impl. both `csrc/xpu/attn/kernel_configs` and `cmake/config/` works for me.

baodii (approved):

YizhouZ (approved):

xinyu-intel (approved):

PR #326, opened 2026-05-06
refactor XpuFusedMoe to reduce host overhead

mayuyuace:


vllm_xpu_kernels/fused_moe_interface.py
```diff
@@ -89,182 +89,182 @@ def process_number(x):
...
+
+        if self.expert_map is not None:
+            self.total_experts_num = self.expert_map.shape[0]
+        else:
+            self.total_experts_num = self.num_experts * self.ep_size
+        self.local_experts_num = self.num_experts
+
+    def apply(
+        self,
+        output,
```
jikunshang, line 190: this will always be inplace, will never return a tensor right?  please add return type for this method

Liangliang-Ma (approved):

vllm_xpu_kernels/fused_moe_interface.py
mayuyuace: Yes, this is a inplace API to keep align with vllm:
https://github.com/vllm-project/vllm/blob/51c1ee9b7c8acbba4899a8ebffd390685d171946/vllm/model_executor/layers/fused_moe/experts/xpu_moe.py#L133

jikunshang: noted. can we still keep this interface to make API compatible? something like:
```
def xpu_fused_moe(...):
     kernel = XpuFusedMoe(...)
     kernel.apply(...)
```
we can remove it in the future.

mayuyuace: The old API has been reverted.
We can remove it at next version.

jikunshang (approved):

PR #332, opened 2026-05-09
benchmark: reduce event recording overhead by using batch timing

xinyu-intel:
## Summary
- Switch standard (e2e) timing from per-iteration event pairs to a single start/end event wrapping the entire measured loop
- Eliminates 2*(N-5) `event.record()` calls from the hot path, removing significant overhead for fast kernels (<50us)
- CalKernelTime providers retain per-iteration events since they pass events into the kernel for sub-kernel measurement

## Test plan
- [x] Verified syntax with `ast.parse` for all modified files
- [x] All pre-commit hooks pass (ruff, codespell, isort, mypy)
- [x] Tested bf16 GEMM benchmark in container — results consistent with expected kernel throughput

### DIFF
--- a/benchmark/benchmark_cutlass_flash_attn_decode.py
+++ b/benchmark/benchmark_cutlass_flash_attn_decode.py
@@ -169,26 +169,18 @@ def benchmark_decode_with_paged_kv(seq_lens, num_heads, head_size, block_size,
           flush=True)
     assert iterations > 5, \
     "Number of iterations should be greater than 5 to account for warmup"
-    total_latency = 0.0
-    ms = 0.0
     queries = [
         torch.rand_like(maybe_quantized_query) for _ in range(iterations)
     ]
 
-    start_events = [
-        torch.xpu.Event(enable_timing=True) for _ in range(iterations - 5)
-    ]
-    end_events = [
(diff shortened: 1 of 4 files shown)


xinyu-intel:
@pengzhao-intel @1pikachu please review. It should reduce the overhead a lot for small shape case.

also, it's align with the xpu-perf test framework. https://github.com/abenmao/xpu-perf/blob/8055226c299074d0dde682c7ed236227b259ec75/micro_perf/backends/INTEL/backend_intel.py#L204-L213

1pikachu:
> @pengzhao-intel @1pikachu please review. It should reduce the overhead a lot for small shape case.
> 
> also, it's align with the xpu-perf test framework. https://github.com/abenmao/xpu-perf/blob/8055226c299074d0dde682c7ed236227b259ec75/micro_perf/backends/INTEL/backend_intel.py#L204-L213

Verified with benchmark tests

1pikachu (approved):

xinyu-intel:
@baodii @jikunshang please review.

also, cc @czhu15

pengzhao-intel (approved):
Thanks for the PR to fix the profiling issue

jikunshang (approved):

PR #337, opened 2026-05-12
Group splitkv for mixed batch decoding 

pengzhao-intel:
# Group Split-K for Mixed-Length Batch Decoding (Xe2)

## 1. Problem statement

The existing `paged_decode_xe2` kernel uses a static `(batch × heads_kv × num_kv_splits)` work-group (WG) grid. For mixed-length decode batches this has two failure modes:

1. **Idle WGs.** A short sequence gets the same `num_kv_splits` slots as a long one. Splits whose tile range is empty exit immediately but still occupy a WG slot, so the launched grid is much larger than the work it carries.
2. **Unfair partition.** `num_blocks_per_split = ceil_div(blocks, splits)` concentrates work on the first `floor(blocks / blocks_per_split)` splits; the trailing split can be tiny or empty.
3. **Hidden GPU→CPU sync.** Any C++-side decision that wants to look at per-seq KV lengths must either pay for a `.item()` round-trip or duplicate the logic the scheduler already needs in Python.

The vLLM scheduler already knows `kv_lens` on the host. This PR pushes the split planning to Python and makes the GPU read a precomputed `work_list`.

## 2. Architecture overview

```
                  vLLM scheduler                  (host)
                         │   kv_lens, num_splits_kv,
                         │   num_xe_cores, num_heads_kv
                         ▼
       flash_attn_interface.build_decode_split_plan      (host, Python)
                         │   splits_per_seq[B]         int32
                         │   work_list[total_wgs, 4]   int32
                         ▼  (.to(device, non_blocking=True))
              flash_attn_varlen_func → varlen_fwd        (PyTorch op)
                         │
                         ▼
              cutlass_paged_decode_impl                  (C++)
                         │  fills args.splits_per_seq,
                         │  args.work_list, args.total_wgs
                         ▼
            DecodeKernelLauncher::run                    (host launch)
                         │  patches scheduler.work_list,
                         │  scheduler.total_wgs,
                         │  scheduler.grid.z = total_wgs × heads_kv
                         ▼
                ┌────────┴────────┐
                ▼                 ▼
      XeFMHAFwdSplitKVKernel    ReduceSplitK
      (compact-grid branch)    (uses splits_per_seq
       reads work_list[wi]      directly, no heuristic
       → seq_idx, tile_start,    reapplied)
         tile_count, split_idx
```

Two execution modes live side-by-side in the kernels:

- **Legacy mode** (`work_list == nullptr`): identical to pre-PR behavior. The scheduler builds a `(B × H_kv × num_kv_splits)` grid and each kernel invocation reapplies the `is_single_split` heuristic on its own.
- **Compact mode** (`work_list != nullptr && total_wgs > 0`): grid is exactly `total_wgs × heads_kv`. Each WG reads one `DecodeWorkItem` from `work_list` and processes exactly its tile range. The kernel does no per-seq split math.

## 3. Algorithm — `build_decode_split_plan`

Inputs: `kv_lens[B]`, `kv_tile`, `num_kv_splits`, `num_xe_cores`, `num_heads_kv`.

```
tiles_per_seq[i]      = max(1, ceil_div(kv_lens[i], kv_tile))
total_tiles           = Σ tiles_per_seq

min_wgs               = max(1, num_xe_cores * 2 / num_heads_kv)
target_tiles_per_wg   = max(4, total_tiles / min_wgs)
min_blocks_for_split  = 32 if kv_tile ≤ 64 else 128

for each seq i:
    if tiles ≤ target_tiles_per_wg  or  tiles < min_blocks_for_split  or  num_kv_splits ≤ 1:
        n_splits = 1                                     # single-split path
    else:
        n_splits = ceil_div(tiles, target_tiles_per_wg)
        n_splits = min(n_splits, num_kv_splits, tiles)   # cap

    base, rem = divmod(tiles, n_splits)
    for s in 0..n_splits-1:
        count = base + (1 if s < rem else 0)
        emit work_item(seq_idx       = i,
                       kv_tile_start = running_offset,
                       kv_tile_count = count,
                       split_idx     = s)
        running_offset += count
```

### Correctness contract

The function returns `(splits_per_seq, work_list)` with these guarantees:

1. `Σ splits_per_seq == work_list.shape[0]`.
2. Every emitted `kv_tile_count ≥ 1`. (Cap `n_splits ≤ tiles` makes `base ≥ 1` regardless of `rem`.)
3. Per seq, the work items partition `[0, tiles)` exactly once (`Σ counts == tiles`, half-open, contiguous).
4. `splits_per_seq[i] ≤ num_kv_splits` so the static reduction buffer `[Oaccum, exp_sums, max_logits]` indexing on the GPU is in-bounds.

Properties 1 and 4 are exactly what `ReduceSplitK` needs to read only the slots that were written. Properties 2 and 3 prevent the "phantom split" problem that motivated the patch.

### Heuristics chosen

- `target_tiles_per_wg = max(4, total_tiles / min_wgs)` keeps the split-K overhead amortized (≥ 4 tiles per launched WG) while still oversubscribing the Xe-core count 2× when work is plentiful.
- `min_blocks_for_split` mirrors the GPU-side single-split threshold so short sequences keep the cheap path.
- Single-split is treated as `is_single_split == true` in the reducer, so the output goes directly to slot 0 with no cross-split reduction.

## 4. C++ plumbing

| File | Change |
| --- | --- |
| `flash_api.cpp` | Schema gains `Tensor? splits_per_seq`, `Tensor? work_list`. Forwarded to both prefill-opt and decode paths of `cutlass_paged_decode_interface`. |
| `attn_interface.{h,cpp}` | Adds the two optional tensors and forwards them to `cutlass_paged_decode_xe2`. |
| `paged_decode_xe2.{h,cpp}` | Wires the tensors into `paged_decode_args_t`: `args.splits_per_seq`, `args.work_list`, `args.total_wgs`. Both default to null/0 (legacy behavior). |
| `paged_decode.hpp` | `paged_decode_args_t` gains `splits_per_seq`, `work_list`, `total_wgs`. `DecodeKernelLauncher::run` patches the scheduler params after `to_underlying_arguments` so the compact-grid path is opt-in per launch. |
| `chunk_prefill_scheduler.hpp` | Adds `DecodeWorkItem` POD struct. `DecodeTileScheduler::Params` gains `work_list`, `total_wgs`. `to_underlying_arguments` shrinks `grid.z` to `total_wgs × heads_kv` when work-list is present. `get_block_coord()` returns a 7-tuple `(blk_q, blk_v, head, idx_b, idx_kv_split, wl_tile_start, wl_tile_count)`; the last two fields are `-1` in legacy mode so the FMHA kernel can branch. |
| `paged_decode_kernel.hpp` | `XeFMHAFwdSplitKVKernel` reads the 7-tuple, branches on `wl_tile_start >= 0`. Compact branch uses the precomputed range. Legacy branch keeps the original heuristic. `ReduceSplitK` reads `splits_per_seq[idx_b]` when provided and trusts it; otherwise mirrors the legacy heuristic. |

The legacy branch is intentionally preserved so:

- existing callers that do not provide `host_kv_lens` keep working,
- prefill (where `max_seqlen_q > 1`) never hits the new path.

## 5. Python interface

`flash_attn_varlen_func` accepts a new optional `host_kv_lens` tensor (or list). When supplied, it is converted once to an `int32` device tensor and used as `seqused_k`. The decode gate is:

```python
block_table is not None
and host_kv_lens is not None
and num_splits_kv is not None and num_splits_kv > 1
and max_seqlen_q == 1
```

Only inside this gate do we build the plan, and we only upload when `work_list_cpu.numel() > 0`. Everywhere else the schema-default `None` tensors are forwarded and the kernel takes the legacy branch.

## 6. Backward compatibility

- `varlen_fwd` schema gained two trailing `Tensor?` arguments; both default to `None`. Callers built against the old signature must be re-bound (this is a kernel package; pinning is expected).
- `paged_decode_args_t` defaults `splits_per_seq=nullptr`, `work_list=nullptr`, `total_wgs=0`. Any C++ caller that does not opt in gets the original grid.
- The FMHA kernel and the reducer both keep their original code paths under `wl_tile_start < 0` / `splits_per_seq == nullptr`.
- `compute_splits_per_seq` (legacy entry point) is kept as a thin shim over `build_decode_split_plan`.

## 7. Performance

Thanks @wuxun-zhang for helping to test.

| row | case               | reference (us) | optimized (us) | delta (us) | perf change | speedup |
| --: | ------------------ | -------------: | -------------: | ---------: | ----------: | ------: |
|  0  | 1,1,4096, (32,2)   |        130.340 |         30.292 |   -100.048 |     +76.76% |  4.303x |
|  1  | 1,1,4096, (32,4)   |        126.117 |         45.559 |    -80.558 |     +63.87% |  2.768x |
|  2  | 1,1,4096, (40,40)  |        217.249 |        222.678 |     +5.429 |      -2.50% |  0.976x |
|  3  | 1,1,4096, (64,8)   |        136.982 |         66.636 |    -70.346 |     +51.35% |  2.056x |
|  4  | 32 x 512, (32,2)   |         71.708 |         69.534 |     -2.174 |      +3.03% |  1.031x |
|  5  | 32 x 512, (32,4)   |        103.518 |         90.393 |    -13.125 |     +12.68% |  1.145x |
|  6  | 32 x 512, (40,40)  |        787.231 |        781.414 |     -5.817 |      +0.74% |  1.007x |
|  7  | 32 x 512, (64,8)   |        166.400 |        163.457 |     -2.943 |      +1.77% |  1.018x |
|  8  | mixed len, (32,2)  |        288.066 |        103.636 |   -184.430 |     +64.02% |  2.780x |
|  9  | mixed len, (32,4)  |        611.125 |        189.415 |   -421.710 |     +69.00% |  3.226x |
| 10  | mixed len, (40,40) |       1649.260 |       1584.079 |    -65.180 |      +3.95% |  1.041x |
| 11  | mixed len, (64,8)  |        385.712 |        349.079 |    -36.632 |      +9.50% |  1.105x |

### DIFF
--- a/csrc/xpu/attn/xe_2/kernel/paged_decode_kernel.hpp
+++ b/csrc/xpu/attn/xe_2/kernel/paged_decode_kernel.hpp
@@ -146,6 +146,7 @@ class XeFMHAFwdSplitKVKernel {
 
     // per-batch mask: true = prefill, false = decode; nullptr = process all
     const bool* is_prefill;
+    const int* splits_per_seq = nullptr;  // per-seq split counts; null => use global num_kv_splits
   };
   using KernelParams = KernelArguments;
 
@@ -251,8 +252,9 @@ class XeFMHAFwdSplitKVKernel {
 
     CUTLASS_PRAGMA_NO_UNROLL
     for (; tile_scheduler.is_valid(); ++tile_scheduler) {
-      auto [blk_q, blk_v, head, idx_b, idx_kv_split] =
-          tile_scheduler.get_block_coord();  // (Q,V,h,b,id_split)
+      auto [blk_q, blk_v, head, idx_b, idx_kv_split,
+           wl_tile_start, wl_tile_count] =
+          tile_scheduler.get_block_coord();  // (Q,V,h,b,split,tile_start,tile_count)
 
       // Skip prefill batches when is_prefill mask is provided
       if (p.is_prefill != nullptr && p.is_prefill[idx_b]) continue;
@@ -328,37 +330,53 @@ class XeFMHAFwdSplitKVKernel {
           make_shape(head_group_q, num_kv_splits, s.num_heads_kv, batch_dim);
       auto shape_sink = make_shape(s.num_heads_kv, head_group_q);
 
-      int num_blocks_per_split =
-          cute::ceil_div(windowed_k_blocks, num_kv_splits);
-
-      // Per-sequence split decision: short sequences are treated as
-      // single-split even when num_kv_splits > 1, avoiding precision
-      // loss from the split-reduce roundtrip.
-      // Scale threshold with tile width: smaller tiles (p64) have fewer
-      // SGs per WG, so splitting becomes beneficial at fewer blocks.
-      constexpr int tile_n = get<1>(TileShapeQK{});
-      constexpr int kMinBlocksForSplit = (tile_n <= 64) ? 32 : 128;
-      bool is_single_split =
-          (num_kv_splits > 1) && (windowed_k_blocks < kMinBlocksForSplit);
-
       int kv_split_offset;
       int num_effective_kv_blocks;
-      if (is_single_split) {
-        // Split 0 processes all blocks; splits 1+ skip entirely.
-        if (idx_kv_split > 0) {
+      int seq_num_kv_splits;
+      bool is_single_split;
+
+      if (wl_tile_start >= 0) {
+        // Compact grid: use pre-computed tile range from work_list.
+        // Python (build_decode_split_plan) has already folded the
+        // single-split heuristic and balanced assignment into the plan.
+        kv_split_offset = k_block0 + wl_tile_start;
+        num_effective_kv_blocks = wl_tile_count;
+        seq_num_kv_splits = (p.splits_per_seq != nullptr)
+            ? p.splits_per_seq[idx_b] : num_kv_splits;
+        is_single_split = (seq_num_kv_splits <= 1);
+      } else {
+        // Legacy path: compute split range on the fly
+        seq_num_kv_splits = (p.splits_per_seq != nullptr)
+            ? p.splits_per_seq[idx_b]
+            : num_kv_splits;
+
+        if (idx_kv_split >= seq_num_kv_splits) {
           continue;
         }
-        kv_split_offset = k_block0;
-        num_effective_kv_blocks = windowed_k_blocks;
-      } else {
-        kv_split_offset = k_block0 + idx_kv_split * num_blocks_per_split;
-        num_effective_kv_blocks = cute::min(
-            windowed_k_blocks - idx_kv_split * num_blocks_per_split,
-            num_blocks_per_split);
+
+        int num_blocks_per_split =
+            cute::ceil_div(windowed_k_blocks, seq_num_kv_splits);
+
+        constexpr int tile_n = get<1>(TileShapeQK{});
+        constexpr int kMinBlocksForSplit = (tile_n <= 64) ? 32 : 128;
+        is_single_split =
+            (seq_num_kv_splits > 1) && (windowed_k_blocks < kMinBlocksForSplit);
+
+        if (is_single_split) {
+          if (idx_kv_split > 0) {
+            continue;
+          }
+          kv_split_offset = k_block0;
+          num_effective_kv_blocks = windowed_k_blocks;
+        } else {
+          kv_split_offset = k_block0 + idx_kv_split * num_blocks_per_split;
+          num_effective_kv_blocks = cute::min(
+              windowed_k_blocks - idx_kv_split * num_blocks_per_split,
+              num_blocks_per_split);
+        }
       }
 
       if (num_effective_kv_blocks <= 0) {
-        // no need computation
         continue;
       }
 
@@ -527,6 +545,7 @@ class ReduceSplitK {
 
     // per-batch mask: true = prefill, false = decode; nullptr = process all
     const bool* is_prefill;
+    const int* splits_per_seq = nullptr;  // per-seq split counts
   };
   using KernelParams = KernelArguments;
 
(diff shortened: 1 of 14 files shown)


csrc/xpu/attn/xe_2/paged_decode_xe2.cpp
```diff
@@ -207,6 +208,70 @@ void cutlass_paged_decode_impl(
       is_varlen ? query.stride(1) : query.stride(1),
       is_varlen ? int64_t{0} : query.stride(0)};
 
+  // Per-sequence adaptive split-K: for batch > 1, compute per-seq splits
+  // to avoid over-splitting short sequences in mixed-length batches.
+  // Only activate for supported GQA configs (ratio >= 8).
+  at::Tensor splits_per_seq_tensor;
+  int num_q_group_size = num_heads_q / num_heads_kv;
+  if (batch_size > 1 && num_kv_splits > 1 && num_q_group_size >= 8) {
```
pengzhao-intel: @copilot  it is not necessary for only GAQ >=8, it can be used in MHA too.

csrc/xpu/attn/xe_2/paged_decode_xe2.cpp
```diff
@@ -207,6 +208,70 @@ void cutlass_paged_decode_impl(
...
+        {batch_size},
+        torch::TensorOptions().dtype(torch::kInt32).device(query.device()));
+
+    auto seqlens_k_ptr = seqlens_k_dev.data_ptr<int>();
+    auto splits_ptr = splits_per_seq_tensor.data_ptr<int>();
+    const int local_window_cap = window_size_left + 1;
+
+    queue.parallel_for(sycl::range<1>(batch_size), [=](sycl::id<1> idx) {
+      int i = static_cast<int>(idx[0]);
+      int seq_k_i = seqlens_k_ptr[i];
```
pengzhao-intel: @xinyu-intel are these data available in the host side?

xinyu-intel: it's not available in the latest attention backend design (CommonMetadata and Flash Attn Backend), all runtime tensors are device tensor so that CUDA Graph as well as Async Scheduling can work well. On the other hand, if the performance benefit is critical, we can consider make a standalone attention backend to achieve. Some other vender implemented backend already did that. https://github.com/vllm-project/vllm/tree/main/vllm/v1/attention/backends

cc @wuxun-zhang @jikunshang

csrc/xpu/attn/xe_2/kernel/paged_decode_kernel.hpp
```diff
@@ -328,23 +329,28 @@ class XeFMHAFwdSplitKVKernel {
           make_shape(head_group_q, num_kv_splits, s.num_heads_kv, batch_dim);
       auto shape_sink = make_shape(s.num_heads_kv, head_group_q);
 
+      // Per-sequence adaptive split count
+      int seq_num_kv_splits = (p.splits_per_seq != nullptr)
+          ? p.splits_per_seq[idx_b]
+          : num_kv_splits;
+
+      // Early exit: this WG's split index exceeds the per-seq allocation
+      if (idx_kv_split >= seq_num_kv_splits) {
```
wuxun-zhang: So here assume static `num_splits` must be larger than max per seq `num_splits`, otherwise will skip part of computation, right?

Also `num_kv_splits` is for allocating `exp_sum` and `max_logits` buffers.

pengzhao-intel: I changed this logic now and only launch the WG with real workload

pengzhao-intel:
@copilot+gpt-5.5  please review

pengzhao-intel:
@copilot resolve the merge conflicts in this pull request

pengzhao-intel:
@xinyu-intel  @wuxun-zhang  @YizhouZ  please reveiw the PR and it's almost done beside CI problem.

pengzhao-intel:
@copilot  resolve conflict

jikunshang (approved):

xinyu-intel (approved):

wuxun-zhang (approved):

PR #342, opened 2026-05-13
Fix: invert can_vectorize branch in per_token_group_quant_8bit_kernel

Yejing-Lai:
Verified the kernel can work fine with vectorization.

### DIFF
--- a/csrc/quantization/fp8/fp8_quant.cpp
+++ b/csrc/quantization/fp8/fp8_quant.cpp
@@ -250,17 +250,17 @@ class per_token_group_quant_8bit_kernel {
 
     const float inverted_scale = 1.0f / (y_s);
     constexpr int VEC_SIZE = 4;
-    if (!can_vectorize) {
+    if (can_vectorize) {
       fp8::ConvertWithScaleOp<true, fp8_type> op{inverted_scale};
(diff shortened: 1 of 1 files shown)


jikunshang (approved):
cc @wuxun-zhang

wuxun-zhang:
minor: add a unit test?

Yejing-Lai:
> minor: add a unit test?

Here is the UT code: https://github.com/vllm-project/vllm-xpu-kernels/blob/main/tests/test_fp8_quant.py#L312

wuxun-zhang:
> > minor: add a unit test?
> 
> Here is the UT code: https://github.com/vllm-project/vllm-xpu-kernels/blob/main/tests/test_fp8_quant.py#L312

Thanks. I suppose this will have performance benefit. Do we have such data?

Yejing-Lai:
> > > minor: add a unit test?
> > 
> > 
> > Here is the UT code: https://github.com/vllm-project/vllm-xpu-kernels/blob/main/tests/test_fp8_quant.py#L312
> 
> Thanks. I suppose this will have performance benefit. Do we have such data?

Tested shape (512, 8064); it will speed up by 1.27x.

xinyu-intel (approved):

PR #344, opened 2026-05-13
[GDN] Accept padded leading dim and slice to num_actual_tokens

### LINKED ISSUE #320: GDN kernel rejects padded inputs from torch.compile / cudagraph capture (incompatible with vLLM PIECEWISE)

jasonboukheir:
## Summary

Fixes #320.

`torch.ops._xpu_C.gdn_attention` previously asserted
`core_attn_out.size(0) == num_actual_tokens` (and the same for
`projected_states_qkvz` / `projected_states_ba`). Under vLLM's
PIECEWISE cudagraph mode the model runner pads inputs to the captured
graph size while `attn_metadata.num_actual_tokens` stays at the real
(unpadded) count, so any decode batch smaller than the captured size
aborts the engine.

This PR relaxes the three `size(0)` checks to `>=` and narrows
`core_attn_out`, `z`, and the two `projected_states` inputs to
`num_actual_tokens` before the kernel launches. Narrowing dim 0 of a
contiguous tensor yields a contiguous view sharing storage, so writes
through the active view land in the caller's buffer at offsets
`[0, num_actual_tokens)`; padded trailing slots are left untouched.

This matches the CUDA path's `_forward_core` in
`vllm/model_executor/layers/mamba/gdn_linear_attn.py:803-808` and
removes the trap for callers without changing the op's signature or
`mutates_args` contract — exact-size callers continue to work
unchanged.

## Test

Added `tests/gdn_attn/test_gdn_attn_padded.py`. Parameterized over
`(num_actual_tokens, padded_size) = [(1, 4), (3, 4), (5, 8)]`. Asserts:

* the op completes without aborting;
* the active prefix `[0:num_actual_tokens)` matches the result of an
  exact-size reference call;
* the padded tail of `core_attn_out` / `z` is left at its zero init
  (kernel doesn't write past the active prefix).

Real-world verification: with the equivalent slicing applied on the
vLLM side as a workaround (jasonboukheir/vllm@ccd77bdf4 — to be
retired by this kernel-side fix), Qwen3.6-35B-A3B + `turboquant_k3v4_nc`
with capture sizes `[1, 4]` lifts from ~20 tok/s (eager) to **65.6
tok/s**; the previously crashing batch-3-padded-to-4 case fired 3,570
times in a stress run with zero errors; 4-concurrent agg = 216.7 tok/s.

## Related

* Issue: vllm-project/vllm-xpu-kernels#320
* vLLM-side workaround that this fix supersedes: vllm-project/vllm#42383
* Similar pattern audit recommended for other XPU custom ops in vLLM's
  `splitting_ops` list (`unified_attention_with_output`, `mamba_mixer`,
  `linear_attention`, `unified_kv_cache_update`, ...). Out of scope for
  this PR.

### DIFF
--- a/csrc/xpu/gdn_attn/gdn_attn_interface.cpp
+++ b/csrc/xpu/gdn_attn/gdn_attn_interface.cpp
@@ -74,26 +74,57 @@ void gdn_attention(
       non_spec_state_indices_tensor.is_contiguous(),
       "non_spec_state_indices_tensor must be contiguous");
 
-  // check core_attn_out shape
-  TORCH_CHECK(core_attn_out.size(0) == num_actual_tokens);
+  // check core_attn_out / z / projected_states_{qkvz,ba} shapes.
+  // Callers running under torch.compile + cudagraph capture pad the leading
+  // dim to the captured graph size while num_actual_tokens stays at the real
+  // (unpadded) count, so accept size(0) >= num_actual_tokens and narrow to the
+  // active prefix below.
+  TORCH_CHECK(
+      core_attn_out.size(0) >= num_actual_tokens,
+      "core_attn_out.size(0) (",
+      core_attn_out.size(0),
+      ") must be >= num_actual_tokens (",
+      num_actual_tokens,
+      ")");
   TORCH_CHECK(core_attn_out.size(1) == num_v_heads / tp_size);
   TORCH_CHECK(core_attn_out.size(2) == head_v_dim);
 
-  // check z shape
   TORCH_CHECK(z.size(0) == core_attn_out.size(0));
   TORCH_CHECK(z.size(1) == core_attn_out.size(1));
   TORCH_CHECK(z.size(2) == core_attn_out.size(2));
 
-  // check projected_states_qkvz shape
-  TORCH_CHECK(projected_states_qkvz.size(0) == num_actual_tokens);
+  TORCH_CHECK(
+      projected_states_qkvz.size(0) >= num_actual_tokens,
+      "projected_states_qkvz.size(0) (",
+      projected_states_qkvz.size(0),
+      ") must be >= num_actual_tokens (",
+      num_actual_tokens,
+      ")");
   TORCH_CHECK(
       projected_states_qkvz.size(1) ==
       num_k_heads / tp_size *
           (2 * head_k_dim + 2 * head_v_dim * num_v_heads / num_k_heads));
 
-  // check projected_states_ba shape
-  TORCH_CHECK(projected_states_ba.size(0) == num_actual_tokens);
+  TORCH_CHECK(
+      projected_states_ba.size(0) >= num_actual_tokens,
+      "projected_states_ba.size(0) (",
+      projected_states_ba.size(0),
+      ") must be >= num_actual_tokens (",
+      num_actual_tokens,
+      ")");
   TORCH_CHECK(projected_states_ba.size(1) == 2 * num_v_heads / tp_size);
+
+  // Narrowing dim 0 of a contiguous tensor yields a contiguous view that
+  // shares storage with the original, so writes through core_attn_out_active
+  // land in the caller's buffer at offsets [0, num_actual_tokens); padded
+  // trailing slots are left untouched (matches the CUDA path in
+  // gdn_linear_attn.py).
+  auto core_attn_out_active = core_attn_out.narrow(0, 0, num_actual_tokens);
+  auto z_active = z.narrow(0, 0, num_actual_tokens);
+  auto projected_states_qkvz_active =
+      projected_states_qkvz.narrow(0, 0, num_actual_tokens);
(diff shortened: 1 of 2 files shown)


xinyu-intel (approved):

jikunshang (approved):
Thanks for your contribution!

PR #350, opened 2026-05-13
fa2 xe2: precompile shared header umbrella for attn_kernels_xe_2

jasonboukheir:
The `chunk_prefill` / `paged_decode` kernel templates share almost their entire transitive include set: `cute`, the CUTLASS-SYCL collective stack, the sycl runtime, `flash_attention_v2` fusion / scheduler / epilogue, the `FMHAKernel` / `PagedDecode` kernel templates. That parse cost is identical across every generated TU and accounts for a non-trivial fraction of per-TU compile wall time on icpx (template-name-lookup over cute / cutlass is template-pathological at this header depth).

## Change

Wire up `target_precompile_headers(attn_kernels_xe_2 PRIVATE attn_pch.hpp)` with an umbrella header that `#include`s both kernel template surfaces. cmake handles the `-Xclang -include-pch -Xclang <build>/cmake_pch.hxx.pch` plumbing automatically for icpx (clang-style PCH), so the per-TU compile commands load the pre-parsed AST snapshot in milliseconds instead of re-parsing.

## Effect

~20–30% wall-time reduction on the FA2 kernel TUs, and the dtype-split PR (#349)'s "more TUs" wall-time penalty disappears entirely. Per-TU peak RSS drops a further 1–2 GB (frontend parse cost is no longer paid per-TU).

## For downstream consumers

Non-cmake consumers (Nix per-TU dynamic-derivations builds) extract the PCH compile command from `compile_commands.json` and realise it as its own derivation; per-TU commands then `-include-pch` the shared PCH artifact. See `attn_pch.hpp` for the icpx PCH compatibility contract.

Draft — minimal cmake change, but I'd appreciate confirmation that upstream CI doesn't have a `-DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON` path that would silently regress.

### DIFF
--- a/csrc/xpu/attn/xe_2/CMakeLists.txt
+++ b/csrc/xpu/attn/xe_2/CMakeLists.txt
@@ -8,3 +8,13 @@ include("paged_decode_configure.cmake")
 paged_decode_configure(paged_decode_kernel_template)
 
 add_xe2_kernel_library(attn_kernels_xe_2 INCLUDE_CMAKE_SOURCE_DIR)
+
+# Shared PCH for the ~3.6k generated kernel TUs. The chunk_prefill +
+# paged_decode kernel templates share almost their entire transitive
+# include set (cute, CUTLASS-SYCL collective stack, sycl runtime); a
+# single PCH collapses that re-parse cost into one drv. See
+# attn_pch.hpp for the umbrella content. cmake handles the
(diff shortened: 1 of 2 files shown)


jasonboukheir:
currently running into some issues, so putting into draft for now.

PR #353, opened 2026-05-15
Change fp8 gemm weight scale shape from [n,k] to [k,n]

Yejing-Lai:
The FP8 gemm weight shape = [k,n]. This PR unifies the weight scale shape with the weight.


jikunshang (approved):
thanks for fixing!

xinyu-intel (approved):

PR #359, opened 2026-05-20
[Benchmark] Adjust FA benchmark configs for B60 memory constraints

1pikachu:
## Summary

- **FA varlen perf**: Reduce Wan2.2 (DiT diffusion) benchmark params to fit Intel B60 (~20G memory):
  - `query_lens`: `"256,512,512,2048"` (was `"1024,2048,2048,8192"`)
  - `kv_lens`: `"256,256,512,2048"` (was `"1024,1024,2048,8192"`)
  - `num_blocks`: `2048` (was `16324`)
- **FA decode perf**: Remove Wan2.2 hardcoded attention shapes (not needed for decode benchmarks).

## Motivation

The Wan2.2 model configs with original params cause OOM on Intel B60 machines (~20G memory). This PR reduces sequence lengths and KV cache blocks for the varlen benchmark to stay within memory limits, and removes the unnecessary decode benchmark case.

### DIFF
--- a/benchmark/src/get_model_config.py
+++ b/benchmark/src/get_model_config.py
@@ -185,12 +185,17 @@ def gen_cutlass_flash_attn_varlen_perf_configs():
                                   is_paged, kv_dtype))
 
         # Add hardcoded attention shapes (diffusion models, etc.)
+        # Use reduced params for B60 (~20G memory):
+        #   smaller seq lens and num_blocks to control memory usage.
+        b60_query_lens = ["256,512,512,2048"]
+        b60_kv_lens = ["256,256,512,2048"]
+        b60_num_blocks = [2048]
         for n_heads, n_kv_heads, h_dim in hardcoded_attn_shapes:
             configs += list(
-                itertools.product(num_seqs, query_lens, kv_lens,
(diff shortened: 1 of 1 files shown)


xinyu-intel (approved):

1pikachu:
@jikunshang Please help review it.

benchmark/src/get_model_config.py
```diff
@@ -185,12 +185,17 @@ def get_configs_from_models():
                                   is_paged, kv_dtype))
 
         # Add hardcoded attention shapes (diffusion models, etc.)
+        # Use reduced params for B60 (~20G memory):
+        #   smaller seq lens and num_blocks to control memory usage.
+        b60_query_lens = ["256,512,512,2048"]
```
jikunshang: should we make it configurable? instead of hardcode B60

1pikachu: Should we automatically detect the machine and modify the config, or control it through parameters?

1pikachu:
@jikunshang  Please help review it, fix the hardcode.

jikunshang (approved):

PR #361, opened 2026-05-21
[Layernorm] optimize rms_norm and fused_rms_norm

zufangzhu:
Purpose:
1. optimize vectorization
2. decrease block size to increase gpu occupacy
3. refactor launch path

Test: 
With qwen3-a3b-30b prefill shape:
rms_norm kernel time: 215.705 us -> 93.94 us
fused_rms_norm_kernel time: 396.457 us -> 310.104 us

|Kernel | Shape | Efficiency|
|-- | -- | --|
rms_norm | [8192, 2048] | 87.6%
rms_norm | [8192, 256] | 93.4%
fused_add_rms_norm | [8192, 2048] | 86.7%

### DIFF
--- a/csrc/layernorm.cpp
+++ b/csrc/layernorm.cpp
@@ -228,8 +228,98 @@ void call_rms_norm_kernel(
   });
 }
 
-template <typename scalar_t>
+template <typename scalar_t, int width>
 class fused_add_rms_norm_kernel {
+ public:
+  fused_add_rms_norm_kernel(
+      scalar_t* __restrict__ input_,     // [..., hidden_size]
+      scalar_t* __restrict__ residual_,  // [..., hidden_size]
+      const int64_t input_stride_,
+      const scalar_t* __restrict__ weight_,  // [hidden_size]
+      const float epsilon_,
+      const int num_tokens_,
+      const int hidden_size_,
+      sycl::local_accessor<float, 1> s_variance_)
+      : input(input_),
+        residual(residual_),
+        input_stride(input_stride_),
+        weight(weight_),
+        epsilon(epsilon_),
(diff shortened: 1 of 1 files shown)


mayuyuace:
Overall LGTM.
Looks like the two kernels still have gap compared with L20.
Will you do more optimization or the two kernels has reach hardware peak?

mayuyuace (approved):

csrc/layernorm.cpp
```diff
@@ -202,34 +245,174 @@ void call_rms_norm_kernel(
   auto out_ptr = out.data_ptr<scalar_t>();
   auto input_ptr = input.data_ptr<scalar_t>();
   auto weight_ptr = weight.data_ptr<scalar_t>();
+
+  const int max_block_size = (num_tokens < 256) ? 1024 : 256;
```
jikunshang, line 249: can you explain why here?

zufangzhu: If num_tokens is big enough, decrease the workgroup size to make more workgroup can be dispatched to gpu and execute simultaneously.

xinyu-intel (approved):

mayuyuace:
I have seen updated efficiency data.
Great!

csrc/layernorm.cpp
jikunshang: maybe we should avoid use such magic number.

jikunshang (approved):

csrc/layernorm.cpp
jikunshang: maybe we should avoid use such magic number.

zufangzhu: I can tune it later.

PR #363, opened 2026-05-21
[XPU][FMHA] Add reverse-odd-heads Q-tile scheduler for causal prefill

pengzhao-intel:
## Motivation

Thanks cong provides the optimization idea.

In FA2 chunk-prefill on Xe2, the default `XeFHMAIndividualTileScheduler` dispatches Q-tiles in the same order for every head. For **causal** prefill this creates a per-head load imbalance:

- Tile `(head=h, q_blk=0)` only attends the diagonal block of KV — light work.
- Tile `(head=h, q_blk=last)` attends nearly the full KV history — heavy work.

Because all heads start from `q_blk=0`, the first wave is uniformly light and the last wave is uniformly heavy. XE cores idle waiting for the heaviest wave to drain.

## Change

Introduce `XeFHMAIndividualReverseOrderTileScheduler`, a drop-in scheduler variant that **reverses the Q-tile order for odd-indexed heads**. Adjacent heads now mix light and heavy tiles in the same wave, smoothing per-wave occupancy.

Selection is per-call via a new bool on the args struct:

```cpp
struct chunk_prefill_args_t {
  ...
  bool reverse_q_order = false;   // new
};
```

The FA2 op sets it from env var `VLLM_XPU_FA_REVERSE_ODD_HEADS`:

| Value | Behavior |
| --- | --- |
| `unset` / `auto` | Shape-based heuristic (default) |
| `1` / `on` / `true` | Force enabled |
| `0` / `off` / `false` | Force disabled |

Auto heuristic:

```cpp
reverse = is_causal && !is_local && !is_sink && num_heads_q >= 2;
```

## Performance evidence

Intel Arc B580, FA2 varlen, MQA q=2 / kv=1, head_size=128, fp16, causal. Numbers are `FlashAttention_Kernel_Time(us)` speedup of `on` over `off`.

### Threshold sweep (bs ∈ {1, 2, 4}, per-seq qlen ∈ [2048, 32768])

| Shape family | Range | Median |
| --- | --- | --- |
| bs=1, qlen 8k–32k | +7 % … +50 % | ~+20 % |
| bs=2, qlen 4k–10k | +8 % … +44 % | ~+27 % |
| bs=4, qlen 2k–4k | -0.7 % … +22 % | ~+15 % |

### Long-context shapes (total kvlen up to 262k)

| Shape | Speedup |
| --- | --- |
| bs=1, qlen 32k–131k (paged & non-paged) | +1.6 % … +9.5 % |
| bs=2 / bs=6, qlen 10k (multi-batch short prefill) | +5 % … +22 % |
| bs=1 / bs=2, qlen 8k + cached_kv 57k | +2 % … +3 % |

Worst observed: -0.7 % on bs=4, qlen=4k non-paged — within run-to-run noise. Every other config across 28 measured shape variants showed a speedup.

## API / backwards compatibility

- `reverse_q_order` defaults to `false`, so callers that construct `chunk_prefill_args_t` without going through fmha_xe2.cpp are unaffected.
- Env-var default is `auto`, i.e. enabled for causal multi-head prefill. To restore strict pre-PR behavior set `VLLM_XPU_FA_REVERSE_ODD_HEADS=off`.

## Testing

- All existing chunk_prefill code paths are unchanged (head_size, block_size, mask, sink, paged, varlen, fp8 KV). The new scheduler is only instantiated when the new bool is true.
- New benchmark configs are automatically included in benchmark_cutlass_flash_attn_varlen.py.
- Recommended pre-merge check: run `pytest tests/flash_attn/test_flash_attn_varlen_func.py` with both `VLLM_XPU_FA_REVERSE_ODD_HEADS=on` and `off` to confirm numeric equivalence.

### DIFF
--- a/benchmark/src/get_model_config.py
+++ b/benchmark/src/get_model_config.py
@@ -235,6 +235,33 @@ def gen_cutlass_flash_attn_varlen_perf_configs():
                 configs.append((1, "1025", "1025", (8, 8), 64, 64, (-1, -1),
                                 out_dtype, None, 2048, 2, None, False, causal,
                                 False, None))
+
+        # MQA(q=2, kv=1) prefill shapes covering chunked-prefill (new tokens +
+        # cached KV) and long-context scenarios. Used to validate the
+        # XeFHMAIndividualReverseOrderTileScheduler variant selected by the
+        # auto-heuristic / VLLM_XPU_FA_REVERSE_ODD_HEADS=on env override.
+        # head_dim=128, fp16, causal. Each tuple is
+        # (bs, q_heads, kv_heads, qlen, cached_kv_len).
+        mqa_prefill_shapes = [
+            (1, 2, 1, 10240, 0),
+            (1, 2, 1, 32768, 0),
+            (1, 2, 1, 8192, 57344),
+            (1, 2, 1, 65536, 65536),
+            (1, 2, 1, 131072, 131072),
+            (2, 2, 1, 10240, 0),
+            (6, 2, 1, 10240, 0),
+            (2, 2, 1, 8192, 57344),
+        ]
+        # num_blocks=16384 fits the largest shape (kvlen=262144 -> 4096 blocks).
+        for bs, qh, kvh, qlen, cached in mqa_prefill_shapes:
+            kvlen = qlen + cached
+            qlens_str = ",".join([str(qlen)] * bs)
+            kvlens_str = ",".join([str(kvlen)] * bs)
+            for paged in [False, True]:
+                configs.append(
+                    (bs, qlens_str, kvlens_str, (qh, kvh), 128, 64, (-1, -1),
+                     torch.float16, None, 16384, 2, None, False, True, paged,
+                     None))
         return configs
 
     configs = get_configs_from_models()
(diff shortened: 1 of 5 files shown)


pengzhao-intel:
@copilot  fix the comment issues

xinyu-intel (approved):

wuxun-zhang (approved):

wuxun-zhang:
I think later we can also try longest processing time scheduling (process longer request firstly from all heads), even for single head, it could help avoid load imbalance issue.

pengzhao-intel:
> I think later we can also try longest processing time scheduling (process longer request firstly from all heads), even for single head, it could help avoid load imbalance issue.

Yes， I think current implementation is only one way and we may provide different scheduler for different situation.

wuxun-zhang (approved):

xinyu-intel (approved):

xinyu-intel:
@jikunshang we may need a more proper way to handle env variables.

baodii (approved):
LGTM. @YizhouZ pls comment.

YizhouZ:
Let's wait for benchmark to finally merge this PR.

pengzhao-intel:
> Let's wait for benchmark to finally merge this PR.

Let me know when the benchmark is done

PR #368, opened 2026-05-25
[XE2] Support MTP of QWEN model

mayuyuace:
Vllm PR: https://github.com/vllm-project/vllm/pull/43565
Tested with Qwen/Qwen3-Next-80B-A3B-Instruct and triton flash attention.

lm_eval results of "num_speculative_tokens": 2:
image

Throughputs on B60, i/o 1024/512:
image

For comparation, throughputs without mtp:
image


jikunshang:
@YangQun1 @wuxun-zhang  can you take a look? it will benefit qwen3.5/3.6 I suppose.

jikunshang (approved):

csrc/xpu/gdn_attn/causal_conv1d.hpp
```diff
@@ -444,9 +468,391 @@ struct update_states_kernel {
...
+#pragma unroll
+          for (int i = 0; i < Width - 1; ++i) {
+#pragma unroll
+            for (int e = 0; e < elems_per_item; ++e) {
+              save_state_ptr[i * conv_elems + reordered_elems_id + e] =
+                  local_input[Width * e + i + 1];
+            }
+          }
+        }
+      }
```
jasonboukheir, line 823: It's true that writing each column 0..K-1 keeps per-acceptance rollback consistent, but FLA's [`_causal_conv1d_update_kernel`](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/mamba/ops/causal_conv1d.py#L838-L856) provides the same rollback property without using K distinct slots: it only ever touches cache_indices[seq, 0], treats that single slot as a length-state_len rolled history, and uses num_accepted_tokens - 1 as a within-slot column offset.

Why prefer the per-K-slot snapshot layout here over FLA's rolled-history-at-slot-0 layout?

xinyu-intel (approved):

csrc/xpu/gdn_attn/causal_conv1d.hpp
mayuyuace: I think both can be right.
I just followed the way ssm states use cache_indices.

mayuyuace: And, I tested the overhead of causal_conv1d_spec_kernel, launching one time only spend about 5 us on B60 if num spec tokens is 2, so I think we do not need to spend much effort to this kernel if functionality is right.

PR #377, opened 2026-05-27
[Benchmark] Add GDN attention benchmark

pengzhao-intel:
### Summary
Add `benchmark/benchmark_gdn_attn.py` for `torch.ops._xpu_C.gdn_attention`. Reports latency, achieved memory bandwidth, MBU and TFLOPS via `triton.testing.perf_report`, suitable for nightly performance tracking.

### Workload coverage
| Mode | Description | Examples |
|---|---|---|
| `decode` | Pure single-token batches (non-spec) | `decode_b{1,8,32,128,256}` |
| `prefill` | Single/multi-request long contexts | `prefill_b1_{1k,4k,8k}`, `b{4_2k,8_1k,16_512,32_256}` |
| `mix` | Mixed prefill+decode at varying decode ratio | `mix_b32_1k_d050`, `b64_512_d075`, `b16_2k_d025` |
| `spec_decode` | Speculative decode, MTP-1/3 | `spec_b{16,32,64}_mtp1`, `spec_b16_mtp3` |

### Shape coverage
| Shape | num_k_heads | num_v_heads | head_k_dim | head_v_dim | Notes |
|---|---|---|---|---|---|
| `Qwen3-Next-80B_tp{1,2,4,8}` | 16 / tp | 32 / tp | 128 | 128 | Production target |
| `Synthetic_MHA_16x16` | 16 | 16 | 128 | 128 | MHA baseline |
| `Synthetic_16x64x128` | 16 | 64 | 128 | 128 | Higher v-head count |
| `Synthetic_16x32x256` | 16 | 32 | 256 | 256 | Wider head_dim |

### Reported metrics
- **GDN(us)** — kernel latency
- **GDN_memBandwidth(GB/s)** — bytes moved / time, lower-bound estimate (Q/K/V/g/β reads + O writes + recurrent state)
- **GDN_MBU(%)** — achieved BW vs. per-device peak from `benchmark/presets.py`
- **GDN_TFLOPS** — conv1d (≈ 2·n_tok·QKV·width) + SSM (≈ 7·nv·hv·hk per token)

### Sample results (Intel Arc B580, peak BW 456 GB/s, peak BF16 117 TFLOPS)

**Qwen3-Next-80B @ tp=1, bf16**

| Workload | Latency (µs) | BW (GB/s) | MBU (%) | TFLOPS |
|---|---:|---:|---:|---:|
| decode_b1 | 23.07 | 88.4 | 19.4 | 0.143 |
| decode_b8 | 88.16 | 208.2 | 45.7 | 0.338 |
| decode_b32 | 327.39 | 215.3 | 47.2 | 0.360 |
| decode_b128 | 1169.75 | 327.6 | 71.8 | 0.569 |
| decode_b256 | 1724.07 | 331.9 | 72.8 | 0.554 |
| prefill_b1_1k | 962.47 | 45.7 | 10.0 | 3.98 |
| prefill_b1_4k | 2580.31 | 66.2 | 14.5 | 5.85 |
| prefill_b1_8k | 5007.62 | 67.7 | 14.9 | 6.15 |
| prefill_b4_2k | 4821.27 | 71.3 | 15.6 | 6.32 |
| prefill_b8_1k | 4991.74 | 69.5 | 15.2 | 6.12 |
| prefill_b16_512 | 5152.27 | 72.6 | 15.9 | 5.94 |
| prefill_b32_256 | 5544.46 | 73.3 | 16.1 | 5.51 |
| mix_b32_1k_d050 | 11953.98 | 62.1 | 13.6 | 5.13 |
| mix_b64_512_d075 | 8161.51 | 59.0 | 12.9 | 3.76 |
| mix_b16_2k_d025 | 15791.63 | 66.1 | 14.5 | 5.79 |
| spec_b16_mtp1 | 197.94 | 186.2 | 40.8 | 0.592 |
| spec_b32_mtp1 | 414.09 | 176.9 | 38.8 | 0.561 |
| spec_b64_mtp1 | 843.07 | 174.3 | 38.2 | 0.578 |
| spec_b16_mtp3 | 441.27 | 85.9 | 18.8 | 0.545 |

### Roofline observation
Arithmetic intensity for all workloads (1.6–90 FLOP/B) sits well below the B580 ridge point (≈256 FLOP/B), so every case is memory-bound. Larger decode batches (b128/b256) approach 70%+ MBU; prefill/mix sit at 10–16% MBU, indicating BW is not yet saturated — pure implementation headroom rather than an architectural ceiling.

### Usage
```bash
python benchmark/benchmark_gdn_attn.py --save-path ./configs/gdn_attn/
```

### Notes
- Hardware presets (`benchmark/presets.py`) need a matching token from `torch.xpu.get_device_name()` for MBU to be reported; otherwise MBU shows `nan`. A `b580` entry can be added in a follow-up.

### DIFF
--- a/benchmark/benchmark_gdn_attn.py
+++ b/benchmark/benchmark_gdn_attn.py
@@ -1,0 +1,500 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+# ruff: noqa: E402
+"""
+Benchmark `torch.ops._xpu_C.gdn_attention` (Gated DeltaNet linear attention,
+used by Qwen3-Next).
+
+Algorithm path is described in csrc/xpu/gdn_attn/gdn_attn_interface.cpp:
+    causal_conv1d(qkv) + gated_delta_rule(SSM recurrence).
+
+Workload modes:
+    - prefill : every sequence is a prefill chunk (varlen)  -> XE2 chunked path
+    - decode  : every sequence is exactly 1 new token       -> native path
+    - mix     : mix of prefill + decode                     -> XE2 chunked path
+    - spec    : MTP / speculative decode (drafts per seq)   -> native path
+
+Output format matches benchmark_cutlass_flash_attn_decode.py
+(triton.testing.perf_report -> nightly picks up the auto-generated CSV).
+"""
+# isort: off
+import gc
+import random
+from dataclasses import dataclass
+
+import torch
+import triton
+import triton.testing
+
+from utils import bootstrap_benchmark_env, ensure_save_path_exists
+
+bootstrap_benchmark_env(__file__)
+
+import vllm_xpu_kernels._xpu_C  # noqa: F401
+from benchmark.presets import get_hardware_preset
+from tests.utils import parse_args, seed_everything
+# isort: on
+
+DEVICE = "xpu"
+
+
+def clear_xpu_cache():
+    torch.xpu.empty_cache()
+    torch.xpu.synchronize()
+    gc.collect()
+
+
+# ----------------------------------------------------------------------------
+# Model shape presets
+# ----------------------------------------------------------------------------
+# Qwen3-Next is the canonical GDN model. The two published variants
+# (80B-A3B-Instruct / Thinking) share the same linear-attention shape:
+# linear_num_key_heads=16, linear_num_value_heads=32,
+# linear_key_head_dim=128, linear_value_head_dim=128, conv_kernel_size=4.
+@dataclass(frozen=True)
+class GdnShape:
+    name: str
+    num_k_heads: int
+    num_v_heads: int
+    head_k_dim: int
+    head_v_dim: int
+    conv_width: int = 4
+    tp_size: int = 1
+
+
+MODEL_SHAPES = [
+    GdnShape("Qwen3-Next-80B_tp1", 16, 32, 128, 128, 4, tp_size=1),
+    GdnShape("Qwen3-Next-80B_tp2", 16, 32, 128, 128, 4, tp_size=2),
+    GdnShape("Qwen3-Next-80B_tp4", 16, 32, 128, 128, 4, tp_size=4),
+    GdnShape("Qwen3-Next-80B_tp8", 16, 32, 128, 128, 4, tp_size=8),
+    # Synthetic shapes to stress wider / larger configurations.
+    GdnShape("Synthetic_MHA_16x16",  16, 16, 128, 128, 4),
+    GdnShape("Synthetic_16x64x128",  16, 64, 128, 128, 4),
+    GdnShape("Synthetic_16x32x256",  16, 32, 256, 256, 4),
+]
+
+
+# ----------------------------------------------------------------------------
+# Workload presets
+# ----------------------------------------------------------------------------
+@dataclass(frozen=True)
+class Workload:
+    name: str
+    mode: str            # "prefill" | "decode" | "mix" | "spec"
+    batch_size: int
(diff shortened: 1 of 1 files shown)


xinyu-intel (approved):

benchmark/benchmark_gdn_attn.py
```diff
@@ -0,0 +1,500 @@
...
+    bytes_out = n_tok * (out_per_tok * 2) * bpe_out     # core_attn_out + z
+    bytes_conv_state = batch * (width - 1) * qkv_per_tok * bpe_in * 2  # RW
+    bytes_ssm_state = batch * nv * hk * hv * bpe_ssm * 2              # RW
+    bytes_weights = qkv_per_tok * width * bpe_in + qkv_per_tok * bpe_in
+
+    return (bytes_in + bytes_out + bytes_conv_state
+            + bytes_ssm_state + bytes_weights)
+
+
+def estimate_flops(kwargs):
```
mayuyuace, line 373: To enable parallelism, two distinct algorithms are employed for the prefill and decode phases. When calculating the TFLOPS and Bandwidth, is the computation performed based on the actual algorithm used?

pengzhao-intel: @copilot  explain how above question

pengzhao-intel: @mayuyuace Good catch — short answer: no, it is path-agnostic, and that does penalize the prefill numbers.

The current estimate_bytes_moved / estimate_flops are an algorithmic lower bound computed from tensor shapes and num_actual_tokens. They count only what any correct implementation must move:
- mixed_qkvz + mixed_ba reads
- core_attn_out + z writes
- conv_state RW (per seq)
- ssm_state RW (per seq)
- conv weights / bias

For decode (native path) the formula is close to truth — there are no auxiliary tensors, so reported MBU (~70%+ at b128/b256) reflects real HW utilization.

For prefill (XE2 chunked path) the formula undercounts. The chunked algorithm allocates and RWs several scratch tensors that the lower bound ignores. Concretely for prefill_b8_1k on Qwen3-Next-80B tp=1 (nk=16, nv=32, hk=hv=128, n_tok=8192, batch=8, bf16):

  A     [nv, n_tok+pad, chunk_size]  ~64 MB  RW
  A_inv (I-A)^-1, same shape         ~64 MB  RW
  W     [nv, n_tok+pad, hk]          ~128 MB RW
  U     [nv, n_tok+pad, hv]          ~128 MB RW
  ---------------------------------------------
  extra total                        ~384 MB

The lower-bound formula reports ~354 MB / 4.99 ms = 71 GB/s (15% MBU). Counting the chunked scratch RW gives ~738 MB / 4.99 ms ≈ 148 GB/s (~32% MBU), which is closer to the actual on-chip traffic.

So the 10–16% prefill MBU in the table is the "ideal-implementation gap", not the "HW gap". I'll send a follow-up that adds a path-aware extension to the estimator (only counts the scratch tensors when num_prefills > 0 and the XE2 path is active) so prefill rows are not unfairly penalized vs. decode/spec. TFLOPS counts the math (conv + SSM recurrence), which is path-invariant, so it stays as-is.

mayuyuace: OK. 
Therefore, this benchmark merely provides a metric indicating the deviation from an ideal standard; it cannot be used as a basis for projections or as a strict requirement for the kernel. This is because I have observed that, even when compared against the L20, the performance gap is not significant.

mayuyuace (approved):

benchmark/benchmark_gdn_attn.py
pengzhao-intel: Let's have an initial data first and then we can change later.

PR #384, opened 2026-06-02
Fuse RMSNorm/SiLU-gate + MXFP8/MXFP4 block-wise quantization patterns

Yejing-Lai:
Follow up [PYTORCHDGQ-8758](https://jira.devtools.intel.com/browse/PYTORCHDGQ-8758).
Add RMSNorm + MX Quant, ResAdd + RMSNorm + MX Quant, SiLU-gate + MX Quant kernel.
All UT pass.

### DIFF
--- a/csrc/ops.h
+++ b/csrc/ops.h
@@ -76,6 +76,13 @@ void silu_and_mul_per_block_quant(
     std::optional<torch::Tensor> scale_ub,
     bool is_scale_transposed);
 
+void silu_and_mul_mxfp4_quant(
+    torch::Tensor& out,
+    torch::Tensor const& input,
+    torch::Tensor& scales,
+    int64_t group_size,
+    double eps);
+
 void mul_and_silu(torch::Tensor& out, torch::Tensor& input);
 
 void gelu_and_mul(torch::Tensor& out, torch::Tensor& input);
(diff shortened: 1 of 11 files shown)


xinyu-intel (approved):

jikunshang (approved):

csrc/ops.h
```diff
@@ -36,7 +36,17 @@ void rms_norm_per_block_quant(
     std::optional<torch::Tensor> scale_ub,
     std::optional<torch::Tensor> residual,
     int64_t group_size,
-    bool is_scale_transposed);
+    bool is_scale_transposed,
+    bool scale_ue8m0);
```
jikunshang, line 40: seems this is not aligned with vllm interface. should we revert? 
https://github.com/vllm-project/vllm/pull/44034

Yejing-Lai: Will this change have any impact?
The newly added scale_ue8m0 is intended to reuse existing code while supporting MXFP8; it defaults to false, and the behavior aligns with upstream theoretically.
BTW, upstream plans to support MXFP8 in the future (https://github.com/vllm-project/vllm/blob/main/tests/compile/passes/test_fusion.py#L322).
Or should we wait until upstream supports this case before changing it on XPU?

PR #387, opened 2026-06-02
Fix RMSNorm work-group size: round down to power of 2

chaojun-zhang:
## Problem

The kernel produces wrong results when the work-group size is not a power of 2 (root cause unknown). The v0.1.9.1 optimization (PR #361) changed the block-size formula from `min(hidden_size, 1024)` to `min(hidden_size/vec_size, max_block_size)`, which yields non-power-of-2 values for common hidden sizes (e.g. 5120/8=640, 7168/8=896) when `num_tokens < 256`.

This was confirmed by 96 failures in the vllm CI on XPU (hidden_size=5120 and 7168, num_tokens=1,8,17,32).

## Fix

Wrap all four block-size expressions in `call_rms_norm_kernel` and `call_fused_add_rms_norm_kernel` with `std::bit_floor((unsigned)...)` to round down to the nearest power of 2.

## Reproducer

See failed vllm IR tests in https://buildkite.com/vllm/intel-ci/builds/3654#019e8147-facd-4077-bc52-3368443d23f5

### DIFF
--- a/csrc/layernorm.cpp
+++ b/csrc/layernorm.cpp
@@ -1,6 +1,7 @@
 #include <sycl/sycl.hpp>
 
 #include <algorithm>
+#include <bit>
 #include <ATen/DeviceGuard.h>
 #include "utils.h"
 #include "dispatch_utils.h"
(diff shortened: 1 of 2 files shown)


zufangzhu:
Is this true? "sycl::reduce_over_group requires a power-of-2 work-group size", should we create a jira for this? @jikunshang

chaojun-zhang:
> Is this true? "sycl::reduce_over_group requires a power-of-2 work-group size", should we create a jira for this? @jikunshang

we can't confirm sycl::reduce_over_group universally requires power-of-2 work-group sizes. What we observed is: the kernel produces wrong results in CI for non-power-of-2 work-group sizes (hidden_size=5120→WG=640, 7168→WG=896), and rounding down to power of 2 fixes it.  
 
Root cause is unknown. A Jira to investigate further would make sense.

zufangzhu (approved):
LGTM

jikunshang:
can we add some ut on kernel side to reproduce and guard in the future?

jikunshang:
This is not root cause. error may come from igc side. @zufangzhu will follow up. thanks for you work.

zufangzhu:
Root cause: vllm-xpu-kernels _C.abi3.so compiled with newer Intel GPU runtime (NEO 26.18.38308.1) produces incorrect results when run on older runtime (NEO 25.48.36300.8)

PR #376, opened 2026-05-27
remove is_xxx from moe class

mayuyuace:
VLLM PR: https://github.com/vllm-project/vllm/pull/43767
Tested below model, results is meaningful.
model=Qwen/Qwen3-30B-A3B
model=Qwen/Qwen2.5-Omni-7B-AWQ
model=openai/gpt-oss-20b
model=Qwen/Qwen3-30B-A3B-FP8
model=INCModel/Qwen3-30B-A3B-Instruct-2507-MXFP8-LLMC

PR #392, opened 2026-06-03
[MoE] Allow reusable workspaces in XPU fused MoE

### LINKED ISSUE #390: XPU fused MoE allocates scratch tensors on every apply call
## Summary

`XpuFusedMoe.apply()` currently allocates several temporary XPU tensors every time the kernel path runs. In decode-heavy MoE inference, especially with small per-step batch sizes, these repeated allocations add avoidable host/runtime overhead and can make the Python wrapper more expensive than necessary.

This is particularly visible for MoE models where `apply()` is called once per MoE layer per decode step. The required scratch shapes are deterministic from the request shape and layer configuration, so callers should be able to reuse preallocated scratch buffers across calls instead of forcing fresh `torch.empty(...)` allocations inside every invocation.

## Relevant code

At current `main` (`dae50e2aa58301d3161c26d15bfd867dc5954c12`), `apply()` exposes no way to pass reusable workspaces and forwards directly to `_apply_kernel`:

https://github.com/vllm-project/vllm-xpu-kernels/blob/dae50e2aa58301d3161c26d15bfd867dc5954c12/vllm_xpu_kernels/fused_moe_interface.py#L402-L416

Inside `_apply_kernel`, the following scratch tensors are allocated unconditionally on each call:

- `remapped_hidden_states`
  https://github.com/vllm-project/vllm-xpu-kernels/blob/dae50e2aa58301d3161c26d15bfd867dc5954c12/vllm_xpu_kernels/fused_moe_interface.py#L457-L460

- `gemm1_output`
  https://github.com/vllm-project/vllm-xpu-kernels/blob/dae50e2aa58301d3161c26d15bfd867dc5954c12/vllm_xpu_kernels/fused_moe_interface.py#L481-L484

- `act_output`
  https://github.com/vllm-project/vllm-xpu-kernels/blob/dae50e2aa58301d3161c26d15bfd867dc5954c12/vllm_xpu_kernels/fused_moe_interface.py#L498-L503

- `gemm2_output`
  https://github.com/vllm-project/vllm-xpu-kernels/blob/dae50e2aa58301d3161c26d15bfd867dc5954c12/vllm_xpu_kernels/fused_moe_interface.py#L505-L508

Some other per-call tensors such as `rows_per_expert` and `unpermuted_row_to_permuted_row` are also allocated each call, but the large activation/GEMM workspaces are the main concern.

The current mini test scope also does not cover EP fused MoE paths, making it easier to miss allocation-path differences in the configurations used by MoE serving:

https://github.com/vllm-project/vllm-xpu-kernels/blob/dae50e2aa58301d3161c26d15bfd867dc5954c12/tests/fused_moe/test_fused_moe.py#L25-L34

## Repro / measurement sketch

A simple allocation-count or timing repro can be written around `XpuFusedMoe.apply()`:

```python
import torch
from vllm_xpu_kernels.fused_moe_interface import XpuFusedMoe

DEVICE = "xpu"
dtype = torch.bfloat16
input_len = 1
hidden_size = 1024
intermediate_size = 2048
num_experts = 16
topk = 1

hidden_states = torch.randn((input_len, hidden_size), device=DEVICE, dtype=dtype)
w13 = torch.randn((num_experts, 2 * intermediate_size, hidden_size), device=DEVICE, dtype=dtype)
w2 = torch.randn((num_experts, hidden_size, intermediate_size), device=DEVICE, dtype=dtype)
scores = torch.randn((input_len, num_experts), device=DEVICE, dtype=torch.float32)
topk_weights, topk_ids = torch.topk(scores, k=topk, dim=-1, sorted=False)

# Match the layout expected by the current test path.
w13 = w13.transpose(-1, -2).contiguous()
w2 = w2.transpose(-1, -2).contiguous()

moe = XpuFusedMoe(
    w13=w13,
    w13_scales=None,
    w13_bias=None,
    w2=w2,
    w2_scales=None,
    w2_bias=None,
    n_experts_per_token=topk,
    activation="silu",
    num_experts=num_experts,
)

out = torch.empty_like(hidden_states)
for _ in range(1000):
    moe.apply(out, hidden_states, topk_weights, topk_ids)
```

Under profiling or allocation tracing, each call creates new scratch buffers for remap, GEMM1, activation, and GEMM2. This creates avoidable allocator churn in decode loops where the shapes are stable.

## Expected behavior

The fused MoE wrapper should allow callers to supply reusable scratch buffers for the large temporary tensors used by the kernel path. The default API should preserve current behavior when no workspaces are provided, but optimized callers should be able to preallocate once and reuse across repeated calls.

Expected properties:

- Existing callers continue to work unchanged.
- Optional workspaces are validated for sufficient capacity.
- Reused workspace output matches the existing allocation path within the same tolerances as current fused MoE tests.
- Workspaces are not reused in a way that aliases live data before it has been consumed.
- EP and quantized fused MoE paths remain covered by tests.

## Suggested tests

1. Add a focused test that compares output from the current allocation path vs. a reusable-workspace path for a small deterministic fused MoE case.
2. Extend mini test parameters to include representative EP fused MoE paths, including int4/MXFP4 variants where applicable.
3. Validate undersized workspace handling, if workspaces are exposed as an API.
4. Run the existing fused MoE test file to ensure the default allocation path remains unchanged.

## Why this matters

For long-running MoE inference, repeated XPU allocations inside each layer call are unnecessary overhead. Allowing workspace reuse would reduce allocator pressure and make the Python wrapper friendlier to high-frequency decode workloads without changing the underlying kernels or default behavior.

ehartford:
## Summary

Fixes #390.

`XpuFusedMoe.apply()` currently allocates several large scratch tensors on every kernel-path call:

- remapped hidden states
- GEMM1 output
- activation output
- GEMM2 output

In decode-heavy MoE inference this creates avoidable allocator churn, since the required workspace shapes are deterministic from `hidden_states.shape`, `n_experts_per_token`, `hidden_size`, and `inter_size`.

This PR adds optional caller-provided workspaces to the XPU fused MoE kernel path. Existing callers keep the current allocation behavior when workspaces are omitted, while optimized callers can preallocate scratch buffers and reuse them across repeated decode steps.

## Changes

- Add optional `workspace13` and `workspace2` parameters to `XpuFusedMoe.apply()` / `_apply_kernel()`.
- Use validated views into those buffers for the large scratch tensors when provided.
- Preserve the existing allocation path when no workspaces are passed.
- Add a regression test comparing reusable-workspace output against the existing allocation path.
- Expand mini fused MoE test coverage to include representative EP variants.
- Add a guard for invalid `num_experts <= 0` in the kernel path.

## Compatibility

The API change is additive. Existing callers do not need to pass any new arguments and should see the same behavior as before.

The workspace buffers are scratch only. The implementation reuses each workspace only after the previous view backed by that workspace has been consumed:

- `workspace13`: remapped hidden states, then activation output
- `workspace2`: GEMM1 output, then GEMM2 output

## Test Plan

Built this branch independently from `origin/main` in a clean worktree with:

```bash
source ~/.venv/bin/activate
PATH=/opt/intel/oneapi/compiler/2026.0/bin:$PATH \
CMPLR_ROOT=/opt/intel/oneapi/compiler/2026.0 \
VLLM_PAGED_DECODE_CONFIG=paged_decode_default.conf \
VLLM_CHUNK_PREFILL_CONFIG=chunk_prefill_default.conf \
python setup.py build_ext --inplace
```

Import smoke:

```bash
source ~/.venv/bin/activate
python - <<'PY'
import vllm_xpu_kernels._C
import vllm_xpu_kernels._xpu_C
import vllm_xpu_kernels._moe_C
print('kernel imports ok')
PY
```

Focused fused MoE tests:

```bash
source ~/.venv/bin/activate
pytest -q tests/fused_moe/test_fused_moe.py --tb=short
```

## Test Result

```text
kernel imports ok
201 passed in 52.92s
```

## Related Work

There is nearby MoE refactor work in #376. This PR is intentionally narrow: it does not refactor the MoE class hierarchy or grouped GEMM signatures; it only adds optional workspace reuse while preserving the default allocation path.

### DIFF
--- a/tests/fused_moe/test_fused_moe.py
+++ b/tests/fused_moe/test_fused_moe.py
@@ -31,6 +31,34 @@ MINI_PYTEST_PARAMS = {
         "dtype": [torch.bfloat16],
         "has_bias": [True],
     },
+    "test_fused_moe_ep": {
+        "m,n,k": [(1, 1024, 1024)],
+        "e": [16],
+        "topk": [1],
+        "ep_rank": [0],
+        "ep_size": [4],
+        "dtype": [torch.float16],
+        "w_dtype": [None],
+        "has_bias": [True],
+    },
+    "test_fused_moe_int4_ep": {
+        "m,n,k": [(1, 1024, 1024)],
+        "e": [16],
+        "topk": [1],
+        "ep_rank": [0],
+        "ep_size": [4],
+        "dtype": [torch.bfloat16],
+        "has_bias": [True],
+    },
+    "test_fused_moe_mxfp4_ep": {
+        "m,n,k": [(1, 1024, 1024)],
+        "e": [16],
+        "topk": [1],
+        "ep_rank": [0],
+        "ep_size": [4],
+        "dtype": [torch.float16],
+        "has_bias": [True],
+    },
 }
 
 
@@ -264,6 +292,70 @@ def test_fused_moe(m, n, k, e, topk, dtype, w_dtype, has_bias):
         atol = 2e-2
     torch.testing.assert_close(output, ref_out, rtol=rtol, atol=atol)
     del fused_moe_impl
+
+
+@torch.inference_mode()
+def test_fused_moe_workspace_reuse_matches_allocation_path():
+    seed_everything(11)
+
+    input_len = 4
+    hidden_size = 128
+    intermediate_size = 256
+    num_experts = 2
+    topk = 1
+    dtype = torch.bfloat16
+
+    a = torch.randn((input_len, hidden_size), device=DEVICE, dtype=dtype) / 16
+    w13 = torch.randn((num_experts, 2 * intermediate_size, hidden_size),
+                      device=DEVICE, dtype=dtype) / 16
+    w2 = torch.randn((num_experts, hidden_size, intermediate_size),
+                     device=DEVICE, dtype=dtype) / 16
+    scores = torch.randn((input_len, num_experts),
+                         device=DEVICE, dtype=torch.float32)
+    expert_scores, expert_indices = torch.topk(scores, k=topk, dim=-1,
+                                               sorted=False)
+
+    w13.data = w13.transpose(-1, -2).contiguous()
+    w2.data = w2.transpose(-1, -2).contiguous()
+    fused_moe_impl = XpuFusedMoe(
+        w13=w13,
+        w13_scales=None,
+        w13_bias=None,
+        w2=w2,
+        w2_scales=None,
+        w2_bias=None,
+        n_experts_per_token=topk,
+        activation="silu",
+        num_experts=num_experts,
+    )
+
+    allocated_output = torch.empty_like(a)
+    fused_moe_impl.apply(
+        output=allocated_output,
+        hidden_states=a,
+        topk_weights=expert_scores,
+        topk_ids=expert_indices,
+    )
+
+    num_moe_inputs = input_len * topk
+    workspace13 = torch.empty((num_moe_inputs,
+                               max(hidden_size, intermediate_size)),
+                              device=DEVICE, dtype=dtype)
+    workspace2 = torch.empty((num_moe_inputs,
+                              max(hidden_size, 2 * intermediate_size)),
+                             device=DEVICE, dtype=dtype)
+    workspace_output = torch.empty_like(a)
+    fused_moe_impl.apply(
+        output=workspace_output,
+        hidden_states=a,
+        topk_weights=expert_scores,
+        topk_ids=expert_indices,
+        workspace13=workspace13,
+        workspace2=workspace2,
+    )
+
+    torch.testing.assert_close(workspace_output, allocated_output,
+                               rtol=2e-2, atol=2e-2)
 
 
 @pytest.mark.parametrize("m,n,k", FUSED_MOE_MNK_FACTORS)
--- a/vllm_xpu_kernels/fused_moe_interface.py
+++ b/vllm_xpu_kernels/fused_moe_interface.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
+import math
 import os
 
 import torch
@@ -405,7 +406,12 @@ class XpuFusedMoe:
         topk_weights,
         topk_ids,
         expert_map=None,
+        workspace13=None,
+        workspace2=None,
     ):
+        # Optional workspaces are scratch buffers sized for the largest view
+        # each buffer backs below. They are reused only after their previous
+        # contents have been consumed.
         if self._use_ref:
             self._apply_ref(output, hidden_states,
                             topk_weights, topk_ids,
@@ -413,7 +419,7 @@ class XpuFusedMoe:
         else:
             self._apply_kernel(output, hidden_states,
                                topk_weights, topk_ids,
-                               expert_map)
+                               expert_map, workspace13, workspace2)
 
     def _apply_ref(
         self,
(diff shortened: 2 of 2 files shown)


mayuyuace:
Num_moe_inputs is a dynamic number.
So I think the code change may raise some problems.
And please provide comparative data both in kernels e2e and models.

xinyu-intel:
@mayuyuace can you investigate whether the memory can be re-used across layers in the single iteration?

mayuyuace:
@xinyu-intel 
Good question!
Without this PR, every layer can re-use memory by allocate and release memory, so total model layers do not need allocate (numbers of layer * single layer memory).
With this PR, every layer would need allocate it's own memory for this.
So, this is indeed something I overlooked: this PR will also lead to higher memory usage.

xinyu-intel:
> @xinyu-intel Good question! Without this PR, every layer can re-use memory by allocate and release memory, so total model layers do not need allocate (numbers of layer * single layer memory). With this PR, every layer would need allocate it's own memory for this. So, this is indeed something I overlooked: this PR will also lead to higher memory usage.

is it possible to allocate/release workspace memory once per iteration instead of per layer?

mayuyuace:
@xinyu-intel 
Torch employs a memory cache mechanism, so there are no actual allocation or release operations on the device; consequently, this only affects the host-side time required for Torch to manage its own memory cache pool.

mayuyuace:
> > @xinyu-intel Good question! Without this PR, every layer can re-use memory by allocate and release memory, so total model layers do not need allocate (numbers of layer * single layer memory). With this PR, every layer would need allocate it's own memory for this. So, this is indeed something I overlooked: this PR will also lead to higher memory usage.
> 
> is it possible to allocate/release workspace memory once per iteration instead of per layer?

I think no.
In a real-world production environment, the input length is a dynamic value; consequently, the sizes of all intermediate tensors cannot be determined upfront. A fixed size can be maintained only during the decoding phase in a testing environment.

xinyu-intel:
> @xinyu-intel Torch employs a memory cache mechanism, so there are no actual allocation or release operations on the device; consequently, this only affects the host-side time required for Torch to manage its own memory cache pool.

you're right, we do see aten::fill host/device call in the trace of each moe layer. Though the device time is very limited, the host call and device bubble cause the decode overhead.

xinyu-intel:
> > > @xinyu-intel Good question! Without this PR, every layer can re-use memory by allocate and release memory, so total model layers do not need allocate (numbers of layer * single layer memory). With this PR, every layer would need allocate it's own memory for this. So, this is indeed something I overlooked: this PR will also lead to higher memory usage.
> > 
> > 
> > is it possible to allocate/release workspace memory once per iteration instead of per layer?
> 
> I think no. In a real-world production environment, the input length is a dynamic value; consequently, the sizes of all intermediate tensors cannot be determined upfront. A fixed size can be maintained only during the decoding phase in a testing environment.

I mean per iteration, a.k.a (num_tokens, hidden_size), it should be known at the beginning of the iteration.

mayuyuace:
> > > > @xinyu-intel Good question! Without this PR, every layer can re-use memory by allocate and release memory, so total model layers do not need allocate (numbers of layer * single layer memory). With this PR, every layer would need allocate it's own memory for this. So, this is indeed something I overlooked: this PR will also lead to higher memory usage.
> > > 
> > > 
> > > is it possible to allocate/release workspace memory once per iteration instead of per layer?
> > 
> > 
> > I think no. In a real-world production environment, the input length is a dynamic value; consequently, the sizes of all intermediate tensors cannot be determined upfront. A fixed size can be maintained only during the decoding phase in a testing environment.
> 
> I mean per iteration, a.k.a (num_tokens, hidden_size), it should be known at the beginning of the iteration.

I see what you mean: you want to allocate a block of memory before each step starts and have every layer reuse it. I think this is feasible, but the change would need to be implemented within vLLM itself; it cannot be achieved solely at the kernels level.

xinyu-intel:
> > > > > @xinyu-intel Good question! Without this PR, every layer can re-use memory by allocate and release memory, so total model layers do not need allocate (numbers of layer * single layer memory). With this PR, every layer would need allocate it's own memory for this. So, this is indeed something I overlooked: this PR will also lead to higher memory usage.
> > > > 
> > > > 
> > > > is it possible to allocate/release workspace memory once per iteration instead of per layer?
> > > 
> > > 
> > > I think no. In a real-world production environment, the input length is a dynamic value; consequently, the sizes of all intermediate tensors cannot be determined upfront. A fixed size can be maintained only during the decoding phase in a testing environment.
> > 
> > 
> > I mean per iteration, a.k.a (num_tokens, hidden_size), it should be known at the beginning of the iteration.
> 
> I see what you mean: you want to allocate a block of memory before each step starts and have every layer reuse it. I think this is feasible, but the change would need to be implemented within vLLM itself; it cannot be achieved solely at the kernels level.

exactly. to check if vLLM modular MOE design can support this.

ehartford:
just try running qwen3.6-35b with and without this patch.  *shrug* not sure what the debate is about.

doopeworld:
Ran this on Arc Pro B70 (BMG-G31) with Qwen3.6-35B-A3B-GPTQ-Int4, batch 1 decode, 40 MoE layers.
Implemented the workspace reuse as a single process-wide cache shared across all layers rather than per-instance, which avoids the memory blowup raised above (per-instance at prefill sizes would need ~40 × 1.5 GB). Only decode-sized buffers are cached; prefill falls through to normal allocation.
Result: no measurable difference. 74.55 tok/s without workspace reuse, 74.20 with (±0.09). PP unchanged at ~10.4k.
This supports @mayuyuace's point that torch's caching allocator already makes the repeated torch.empty calls cheap on device — the churn is real but it isn't costing decode time on this path. The host-side cost that @xinyu-intel observed in the trace appears to be elsewhere.
Happy to re-run on other shapes if useful.

PR #398, opened 2026-06-05
Support fp8 block gemm with fp8_e8m0 scale

Yejing-Lai:
Support FP8 block GEMM with fp8_e8m0 scale and add related UT.
Clean the code that if-else path is too long.


jikunshang (approved):
LGTM. thanks for fixing.
cc @wuxun-zhang

wuxun-zhang (approved):

Yejing-Lai:
@zufangzhu Please review~

zufangzhu (approved):
LGTM.   BTW, do we need to update w8a16 as well?

Yejing-Lai:
> LGTM. BTW, do we need to update w8a16 as well?

w8a16 not include MXFP8 path. It can be kept as it is.

baodii (approved):

PR #399, opened 2026-06-05
[XPU][GDN] Fix tiled prefill conv1d dropping z for GQA ratio 3

jasonboukheir:
## Summary

The tiled prefill conv1d kernel writes the `z` gate with a fixed pool of work items that can only cover **256 values per token in a single pass**. When a model needs more than 256 `z` values per token, everything past the first 256 is **never written** — that part of the gate is left as garbage. Qwen3.6-27B hits this, and the corrupted gate breaks the attention output so generation collapses into a repeated nonsense token.

## Root cause

`z` has `head_v_dim * num_v_heads / num_k_heads` values per token. The z-reorder step in `chunk_causal_conv1d_tiled_xe2` runs **one** pass of 64 work items × 4 values = 256, and only writes those first 256:

- Most configs keep this `<= 256` — e.g. `head_v_dim=128` with a 2:1 value/key head ratio gives exactly 256 — so the single pass covers everything and the bug is invisible.
- Qwen3.6-27B has a **3:1 ratio** (`num_v_heads=48, num_k_heads=16, head_v_dim=128`) → **384 values per token**. The work items only reach the first 256, so the last 128 (the 3rd value-head of each group) are never computed and `z` is left zero/garbage there.

That corrupt `z` flows into `core_attn_out`, so any prefill of `>= conv1d_tile_size` (8) tokens — i.e. the tiled path — produces broken output and the model degenerates into repeated tokens.

## Fix

Loop the z-reorder in steps of 256 until **all** values are written, instead of a single 256-wide pass. When there are `<= 256` values this is identical to the old code; above that it now covers the rest.

## Test

Adds a regression test with a 3:1 head ratio (384 z values per token) on a single short prefill, comparing `z` and `core_attn_out` against the reference. It fails on the current kernel (the 3rd value-head of each group is wrong) and passes with the fix. The existing tests only use the 2:1 ratio, which is why this was never caught.

## Manual Testing

Manually testing on my home server here with an Arc B70 Pro with open webui. Before I was getting !!!!!! characters, now I get full output.

Server configs: https://github.com/jasonboukheir/dotfiles/blob/4c2bb3721b38a86372b23975b649a51b64986af7/hosts/brutus/services/vllm-xpu.nix#L40

My vllm fork:
https://github.com/jasonboukheir/vllm-xpu-nix

### DIFF
--- a/csrc/xpu/gdn_attn/xe_2/chunk_causal_conv1d_tiled_xe2.hpp
+++ b/csrc/xpu/gdn_attn/xe_2/chunk_causal_conv1d_tiled_xe2.hpp
@@ -500,10 +500,14 @@ struct chunk_causal_conv1d_tiled_kernel {
         int out_token_id = pre_chunks * chunk_size_xe2 + token_in_seq;
         int global_tok = lookup(seq_start + token_in_seq);
 
-        // z reorder: each item handles elems_per_item z features
-        // z_dim = 256 for Qwen, items 0-63 cover 64*4=256 features exactly
-        if (local_feat < z_dim) {
-          int z_dim_id = local_feat;
+        // z reorder: each item handles elems_per_item z features per pass.
+        // The 64 items cover feats_per_wg (256) features per pass, so when
+        // z_dim exceeds 256 -- e.g. GQA ratio num_v_heads/num_k_heads == 3
+        // gives z_dim = head_v_dim * 3 = 384 for Qwen3.6 -- a single pass would
+        // drop the tail z features (the 3rd v-head of each k-group). Stride by
+        // feats_per_wg so every z feature is written for any ratio.
+        for (int z_dim_id = local_feat; z_dim_id < z_dim;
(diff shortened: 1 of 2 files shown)


jasonboukheir:
@jikunshang 

Not sure this is actually a bug fix, comment in the previous code made it look like it was intentionally only targeting 256 size. Definitely lmk if there is a better approach! Thanks for reading :smile:

jikunshang:
@YangQun1 @pengzhao-intel @mayuyuace please take a review, thanks!

YangQun1 (approved):
LGTM

mayuyuace:
@pengzhao-intel 
Please help review this PR.

pengzhao-intel (approved):
Thanks to fix the bug :)

pengzhao-intel:
@jikunshang please help reveiw and approve :)

jikunshang (approved):

PR #401, opened 2026-06-08
[XPU][MOE] Optimize XPU MoE activation by skipping invalid rows

ranzhejiang:
Summary
This PR optimizes the XPU fused MoE activation path for EP prefill by passing the valid routed row count into fusion and skipping invalid trailing rows on device.

In EP, the real routed rows can be smaller than num_rows * topk. Without this information, activation still runs on the full upper-bound shape even though grouped GEMM already uses the real routed rows. Carrying the valid row count into fusion avoids redundant activation work without introducing Python-side host synchronization.

This path is only enabled for EP prefill. Decode keeps the existing behavior, and TP is unaffected because it does not use this EP-specific valid-row path

In https://github.com/Tencent/hpc-ops, they also use valid_row_range_ptr to skip unused rows

### DIFF
--- a/csrc/activation.cpp
+++ b/csrc/activation.cpp
@@ -92,12 +92,17 @@ class act_kernel {
   act_kernel(
       scalar_t* __restrict__ out,          // [..., d]
       const scalar_t* __restrict__ input,  // [..., d]
-      const int d)
-      : out_(out), input_(input), d_(d) {}
+      const int d,
+      const int64_t* __restrict__ valid_rows = nullptr)
+      : out_(out), input_(input), d_(d), valid_rows_(valid_rows) {}
 
   void operator() [[sycl::reqd_sub_group_size(32)]] (
       const sycl::nd_item<3>& item_ct1) const {
     const int64_t token_idx = item_ct1.get_group(2);
+    if (valid_rows_ != nullptr &&
+        token_idx >= static_cast<int64_t>(*valid_rows_)) {
+      return;
+    }
     for (int64_t idx = item_ct1.get_local_id(2); idx < d_;
          idx += item_ct1.get_local_range(2)) {
       const scalar_t x = input_[token_idx * d_ + idx];
@@ -109,6 +114,7 @@ class act_kernel {
   scalar_t* __restrict__ out_;          // [..., d]
   const scalar_t* __restrict__ input_;  // [..., d]
   const int d_;
+  const int64_t* __restrict__ valid_rows_;
 };
 
 template <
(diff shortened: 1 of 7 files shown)


xinyu-intel:
fix pre-commit

mayuyuace:
I this if this code change is merged, modes other than EP mode will result in a performance decline, because other modes don't need this.
Please provide specific model execution data including TP only, EP and other mode.

ranzhejiang:
> I this if this code change is merged, modes other than EP mode will result in a performance decline, because other modes don't need this. Please provide specific model execution data including TP only, EP and other mode.

ok，I will test it with performance data

ranzhejiang:
> I this if this code change is merged, modes other than EP mode will result in a performance decline, because other modes don't need this. Please provide specific model execution data including TP only, EP and other mode.

I add `_should_use_ep_valid_tokens` to ensure the optimization only takes effect in **EP prefill** and does not affect **TP**.

| Metric | Before | After | Improvement |
|---|---:|---:|---:|
| Mean TTFT (ms) | 366.98 | 356.65 | 2.8% |
| Median TTFT (ms) | 364.10 | 354.67 | 2.6% |
| P99 TTFT (ms) | 373.97 | 363.29 | 2.9% |

Tested with **Qwen3-30B-A3B** using `--random-input-len 4096 --random-output-len 128`, eager, TP4 with expert parallel, XPU is B60.

In theory, the time savings should become more significant as **EP increases** and **sequence length becomes longer**.

mayuyuace:
Please fix pre-commit.

mayuyuace:
BTW, can you please provide the performance data with tp=4 and without ep? 
Although I think now this PR will not affect pure TP performance.

ranzhejiang:
> BTW, can you please provide the performance data with tp=4 and without ep? Although I think now this PR will not affect pure TP performance.



> BTW, can you please provide the performance data with tp=4 and without ep? Although I think now this PR will not affect pure TP performance.

| Metric | Before | After | Improvement |
|---|---:|---:|---:|
| Mean TTFT (ms) | 359.50 | 358.68 | 0.2% |
| Median TTFT (ms) | 358.02 | 356.57 | 0.4% |
| P99 TTFT (ms) | 365.64 | 365.70 | -0.0% |

Tested with **Qwen3-30B-A3B** using `--random-input-len 4096 --random-output-len 128`, eager,  pure TP4 without expert parallel, XPU is B60.

mayuyuace (approved):

mayuyuace:
@Liangliang-Ma 
Please help review.

PR #379, opened 2026-05-29
[XPU][GDN] Optimize l2_norm and chunk_fwd_o kernel for GDN attention prefill path

YangQun1:
# Essential Elements of an Effective PR Description Checklist

- [x] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [x] The test plan, such as providing test command.
- [x] The test results, such as pasting the results comparison before and after, or e2e results
- [x] (Optional) The necessary documentation update, such as updating `supported_models.md` and `examples` for a new model.

PLEASE FILL IN THE PR DESCRIPTION HERE ENSURING ALL CHECKLIST ITEMS ABOVE HAVE BEEN CONSIDERED.

## Purpose

Optimize the l2norm and chunk_fwd_o kernel performance in GDN attention prefill path
- fuse l2norm to conv1d as an epilogue, and give a optimized fallback for can't fused cases
- optimize chunk_fwd_o kernel with register data reuse

## Test Plan

 - [x] Functional correctness verified against baseline (same output from gdn_attention)
 - [x] Performance validation with benchmrk_gdn_attn.py scripts (200 iterations)

## Test Result

All tests/benchmarks run on B60

- [x] All existing unit tests passed
- [x] Performance improvement for prefill path

| shape_name | workload_name | dtype_str | latest_main_GDN(us) | this_PR_GDN(us) | speedup_ratio |
| --- | --- | --- | --- | --- | --- |
| Qwen3-Next-80B_tp1 | prefill_b1_1k | bf16 | 1143.744533 | 1046.057867 | 1.093386 |
| Qwen3-Next-80B_tp1 | prefill_b1_4k | bf16 | 2773.193867 | 2361.693867 | 1.174239 |
| Qwen3-Next-80B_tp1 | prefill_b1_8k | bf16 | 5320.065067 | 4474.919200 | 1.188863 |
| Qwen3-Next-80B_tp1 | prefill_b4_2k | bf16 | 5247.928267 | 4273.057067 | 1.228144 |
| Qwen3-Next-80B_tp1 | prefill_b8_1k | bf16 | 5300.384533 | 4370.102133 | 1.212874 |
| Qwen3-Next-80B_tp1 | prefill_b16_512 | bf16 | 4870.897600 | 4372.238133 | 1.114051 |
| Qwen3-Next-80B_tp1 | prefill_b32_256 | bf16 | 5050.569600 | 4479.063733 | 1.127595 |
[...]

PR #411, opened 2026-06-12
Fix shared-local-memory write-after-read race in GDN chunk kernels (NaN at >=32K on XE2)

alankessler:
### Environment

main (6bf5660; same race on v0.1.8/v0.1.9), torch 2.12.0+xpu, oneAPI 2025.3, IGC
2.28.4, Arc Pro B70 (BMG-G31), compute-runtime 26.05, kernel 6.17.

### Summary

I repeatedly see GDN models at >=32K on XE2 emit NaN logits ('!!!!') or
nondeterministic output at temperature 0.

`chunk_compute_A_kernel` and `chunk_compute_wu_kernel` refill `g_slm_ptr` and
`beta_slm_ptr` each (chunk, v_head) iteration with only the post-fill barrier. Nothing
orders the refill against the previous iteration's readers, so a sub-group that
finishes its GEMM early overwrites the SLM while slower sub-groups still read it in the
exp() epilogue (`sycl::exp(g_slm_ptr[m_idx] - g_slm_ptr[n_idx])`). Mismatched cumsum
pairs give wrong decay factors: a low-bit error when stale and fresh are close,
and an overflow to NaN when they're far apart. The corruption lands in one (chunk, v_head) cell and
propagates through ssm_state, matching the signature: NaN in 1 of `num_v_heads`,
mid-sequence onset.

This is the "random error" behind the `num_actual_tokens == 8192` skips in
`tests/gdn_attn/test_gdn_attn.py`, added in #261. The fix makes them pass; patch 2
removes them.

### Reproduction

I ran both reproductions on main (6bf5660), BMG-G31, stock vs fixed.

With the 8192 skips removed, the repo's own test (bf16 prefill/mix-mode):

| kernels | 16 cases x 2 reps |
|---|---|
| stock | 7/16, 8/16 FAILED |
| + fix | 16/16, 16/16 passed |

Then `repro_gdn_inverse_race.py`, which drives `torch.ops._xpu_C.gdn_attention`
directly in chains of 16 state-linked 8192-token prefills (NK=16 HK=128 NV=48 HV=128),
counting NaN/all-zero output:

| kernels | NaN out of 2000 |
|---|---|
| stock | 1/2000 |
| + fix | 0/2000 |

I used the same harness on v0.1.9 (torch 2.11): 3/2000 to 0/2000, and bitwise replay
81% to 0%. Torch 2.12 seems to introduce a separate nondeterminism: on main, replay is
95-99% on both stock and fixed, so I am reporting only NaN results there.

I also forced L1-bypass loads, which made it worse on v0.1.9 (14/2000), not better.
Slower readers widen the refill window, which rules out cache coherence and confirms
the reader-latency-scaled WAR.

### Fix

I add three `local_space` barriers, one before each SLM refill, in
`chunk_compute_A_kernel`, `chunk_compute_wu_kernel`, `chunk_fwd_o_kernel`. No new
memory, no algorithm change, and no measurable perf cost (decode 22.6 vs 22.9 tok/s,
prefill 3.40 vs 3.41 s, identical greedy output).

Separately, `chunk_fwd_o_kernel` stages tiles through global memory across sub-groups
with only `local_space` fences, which don't order global stores. On v0.1.9 I saw no
behavioral change from this (the per-Xe-core L1 seems to mask it), so the SLM fences
suffice. It's still a data race under the memory model, and #379 changed that staging,
so I can add those fences here if you want them.

### Test

Patch 2 removes the two `num_actual_tokens == 8192` FIXME skips and restores the
core_attn_out and ssm_state assertions (ssm_state was dead code after the skip). On the
repo's shapes (NV=32) the 8192 bf16 cases fail 7-8/16 on stock and pass 16/16 fixed.

### DIFF
--- a/csrc/xpu/gdn_attn/xe_2/chunk_gated_delta_rule_kernels_xe2.hpp
+++ b/csrc/xpu/gdn_attn/xe_2/chunk_gated_delta_rule_kernels_xe2.hpp
@@ -199,6 +199,11 @@ CUTE_DEVICE void chunk_compute_A_kernel(
       const int chunk_start_offset = chunk_id * chunk_size;
 
       for (int v_head_id = 0; v_head_id < num_v_heads; ++v_head_id) {
+        // WAR fence: sub-groups from the previous (chunk, v_head)
+        // iteration may still be reading g_slm_ptr in the exp()
+        // epilogue; all must arrive before any sub-group refills the
+        // SLM (g_slm_ptr) for this iteration.
+        item.barrier(sycl::access::fence_space::local_space);
         CUTE_UNROLL
         for (int e = local_id; e < chunk_size; e += local_range) {
           g_slm_ptr[e] =
@@ -769,6 +774,10 @@ CUTE_DEVICE void chunk_compute_wu_kernel(
       const int chunk_start_offset = chunk_id * chunk_size;
 
       for (int v_head_id = 0; v_head_id < num_v_heads; ++v_head_id) {
+        // WAR fence: previous iteration's sub-groups may still be
+        // reading beta_slm_ptr/g_slm_ptr; all must arrive before any
+        // sub-group refills the SLM for this iteration.
+        item.barrier(sycl::access::fence_space::local_space);
         CUTE_UNROLL
         for (int e = local_id; e < chunk_size; e += local_range) {
           float beta_value =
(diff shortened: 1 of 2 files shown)


mayuyuace (approved):

mayuyuace:
Thanks for your fixing.
@jikunshang 
Let us observe whether random errors continue to occur after this PR.

jikunshang:
@YangQun1 @pengzhao-intel  PTAL.

YangQun1 (approved):

pengzhao-intel (approved):
thanks to fix the bug

PR #429, opened 2026-06-19
Make XPU flash cache writes stride-aware

LucasWilkinson:
## Summary
- keep the existing vectorized reshape_and_cache_flash path for compact K/V cache rows
- use per-head destination indexing when the cache view has a larger head stride
- fixes writes into content-dim packed FlashAttention K/V views, where K/V heads are separated by 2 * head_size stride

## Context
This came up while validating vllm-project/vllm#44455, which packs K/V into the content dimension. The XPU FlashAttention read path already consumes cache strides, but reshape_and_cache_flash was bulk-copying num_heads * head_size contiguous elements for each token row, which corrupts packed K/V split views.

## Tests
- git diff --check origin/main...HEAD -- csrc/cache.cpp
- /home/LucasWilkinson/local/vllm2/.venv/bin/pre-commit run clang-format --files csrc/cache.cpp

AI assistance was used to inspect the vLLM failure mode and draft this patch.

### DIFF
--- a/csrc/cache.cpp
+++ b/csrc/cache.cpp
@@ -169,10 +169,32 @@ class reshape_and_cache_flash_kernel {
 
     fp8::CopyWithScaleOp<cache_t, scalar_t, kv_dt> k_op{k_scale_val};
     fp8::CopyWithScaleOp<cache_t, scalar_t, kv_dt> v_op{v_scale_val};
-    vectorize_with_alignment<VEC_SIZE>(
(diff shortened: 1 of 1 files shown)


xinyu-intel (approved):
LGTM. Thx for the fix:)

jikunshang:
thanks for fixing. we have a performant impl in #432, and released in vllm-xpu-kernels 0.1.10.1, close this.

Issue #438, opened 2026-06-29
chunk_prepare_kernel: v_head_id can exceed num_v_heads (unguarded) → OOB write → NaN runaway on XE2


TSUMUGI-XE:
## Summary

In the GDN chunk kernels, `chunk_prepare_kernel` (`chunk_gated_delta_rule_kernels_xe2.hpp`) derives the
v-head index from the flat sub-group id without an upper-bound guard:

```cpp
int total_sg_range   = group_range * sg_range;        // total sub-groups launched (hardware-dependent)
const int chunk_range = total_sg_range / num_v_heads; // integer division
const int v_head_id    = total_sg_id / chunk_range;    // <-- can be >= num_v_heads, no guard
...
const_cast<float*>(a)[(chunk_start_offset + sg_local_id * local_num + c)
                      + v_head_id * total_virtual_seqlen] = g_local_sum;   // a is [num_v_heads, total_virtual_seqlen]
```

When `total_sg_range` is **not a multiple of `num_v_heads`**, the highest sub-groups compute
`v_head_id == num_v_heads` (and, when `chunk_range < num_v_heads`, even `num_v_heads + 1`). Those
sub-groups then write `a[v_head_id * total_virtual_seqlen + ...]`, i.e. **one or more head-rows past
the end of the `a` buffer** (`a` has only `num_v_heads` rows). This is an out-of-bounds device write
that lands in whatever allocation follows `a` — in our case the next module's quantization `scales`
buffer — corrupting it to NaN, which then propagates to all downstream outputs (persistent `!!!!` /
all-non-alphanumeric runaway; recoverable only by restart).

## Why this is reachable

- `total_sg_range = group_range * sg_range` is hardware-determined (≈ SM_count × sub-groups-per-WG)
  and need not be a multiple of `num_v_heads`. Example on Arc Pro B70 (BMG-G31), Qwen3.6-27B
  (`num_v_heads = 48`): `total_sg_range ≈ 1280`, `chunk_range = floor(1280/48) = 26`, so the top
  sub-groups get `v_head_id = floor(1279/26) = 49` — two head-rows out of bounds.
- The OOB sub-groups only *write* when they have chunk work (`chunk_id < cumsum_chunks`), so the
  corruption is **load-dependent** (shows up under high-volume continuous batching, where
  `cumsum_chunks` is large), which is why it presents as an intermittent runaway rather than an
  immediate failure.
[...]

PR #439, opened 2026-06-29
Guard v_head_id in chunk_prepare_kernel to prevent OOB write (#438)

TSUMUGI-XE:
Adds the missing `v_head_id >= num_v_heads` guard in `chunk_prepare_kernel`, fixing the out-of-bounds device write reported in #438.

`v_head_id = total_sg_id / chunk_range` (with `chunk_range = total_sg_range / num_v_heads`, integer division) can reach `num_v_heads` when `total_sg_range` is not a multiple of `num_v_heads`, and the unguarded write `a[v_head_id * total_virtual_seqlen + ...]` then lands one or more head-rows past the end of the `a` buffer (`[num_v_heads, total_virtual_seqlen]`), corrupting the adjacent allocation (observed as NaN propagating to all outputs on Arc B70 / Qwen3.6-27B under high-volume continuous batching).

All other chunk sub-kernels already bound `v_head_id` to `[0, num_v_heads)` (for-loop, `% num_v_heads`, or launch dim). This restores the same bound for ChunkPrepare. `v_head_id` is constant across a sub-group, so the early return is taken uniformly (no divergence across the `inclusive_scan_over_group`).

**Verification** (Arc B70 / Qwen3.6-27B-AWQ): before the guard, high-volume bursts corrupted the victim buffer within 1-2 runs; with the guard, sustained bursts ran clean (~32k forwards, ~337k post-sub-kernel NaN scans, 0 corruption). See #438 for the full localization.

### DIFF
--- a/csrc/xpu/gdn_attn/xe_2/chunk_gated_delta_rule_kernels_xe2.hpp
+++ b/csrc/xpu/gdn_attn/xe_2/chunk_gated_delta_rule_kernels_xe2.hpp
@@ -74,6 +74,10 @@ CUTE_DEVICE void chunk_prepare_kernel(
   const int chunk_range = total_sg_range / num_v_heads;
   int chunk_id = total_sg_id % chunk_range;
   const int v_head_id = total_sg_id / chunk_range;
+  // When total_sg_range is not a multiple of num_v_heads, the top sub-group(s)
+  // get v_head_id >= num_v_heads and would write past the end of a[num_v_heads,
+  // ...] (OOB).
+  if (v_head_id >= num_v_heads) return;
 
   const float A_log_exp_h = -sycl::exp(A_log[v_head_id]);
   const float dt_bias_h = static_cast<float>(dt_bias[v_head_id]);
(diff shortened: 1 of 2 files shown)


mayuyuace:
LGTM.
Can you removed the added long comments and add UT for this bug?
BTW, please Signed-off-by to pass CI.

TSUMUGI-XE:
Thanks for the review! Addressed all three:

1. **Trimmed the comment** — the in-kernel note is down to two lines; the full derivation stays in the commit message.
2. **Added a UT** — `tests/gdn_attn/test_gdn_attn.py::test_chunk_prepare_vhead_oob_guard`. It covers the `num_v_heads = 48` (Qwen3.6-27B, GQA ratio 3) shape that drives `v_head_id` to `num_v_heads`: on Battlemage sm_count = 32 sub-slices and sg_range = 32, so `total_sg_range = 1024` isn't a multiple of 48 and the top sub-groups reach `v_head_id = 48`. The existing tests only cover ratio 2 (num_v_heads = 32), which always divides total_sg_range. Passes for fp16/bf16 on Arc B70.
3. **Signed-off-by** added (amended + force-pushed), so DCO should be green now.

mayuyuace:
Please fix pre-commit error.

mayuyuace (approved):

jikunshang (approved):

PR #448, opened 2026-07-03
[Dont merge]Enale FusedMoe kernel V2

yihuaxu:
To optimize the performance for fused_moe

## Purpose
Try to improve the performance based on current split moe kernels to save the io cost.

## Test Result
UT:
Decode:
image

image

Prefill:
image


mayuyuace:
What's your motivation?

mayuyuace:
The sole benefit of this PR is the reduction of intermediate variable movement.
For the prefill phase, performance is primarily compute-bound, so the bandwidth consumed by moving intermediate variables does not impact overall performance.
For the decoding phase, performance is primarily bound by MoE weight transfers, making the overhead of moving intermediate variables virtually negligible.
I don't think this PR will lead to a significant overall improvement in the model.

yihuaxu:
> What's your motivation?

To optimize the performance based on current MOE implementation.

yihuaxu:
> The sole benefit of this PR is the reduction of intermediate variable movement. For the prefill phase, performance is primarily compute-bound, so the bandwidth consumed by moving intermediate variables does not impact overall performance. For the decoding phase, performance is primarily bound by MoE weight transfers, making the overhead of moving intermediate variables virtually negligible. I don't think this PR will lead to a significant overall improvement in the model.

Please refer to the test result.

mayuyuace:
According your data, I calculated your kernel performance:
image
Looks like my kernel has poor performance, please compare with hardware peak performance instead of mine.
I will finetune the split kernel.
Thanks.

PR #455, opened 2026-07-09
[ATTN][DECODE] Route small uniform-`q_len` causal attention to the split-K decode path

baodii:
## Summary

In speculative decoding, every sequence in a batch issues the same small number
of query tokens (`q_len = num_speculative_tokens + 1`, typically 2/3/4). Because
`max_seqlen_q > 1`, these batches were dispatched to the **chunk-prefill**
kernel, which has no split-K parallelism over the KV dimension. With a long KV
cache and only a handful of query rows, the GPU is badly under-utilized, so the
prefill path is slow.

This PR adds a fast path in `flash_attn_varlen_func` that redirects small,
**uniform** query-length, causal, paged batches to the split-K **decode**
kernel, which already parallelizes across the KV dimension. The redirect is
exact and runs in a **single** `varlen_fwd` launch.

## Approach

For a uniform-`q_len` batch, each `(sequence, query-position)` pair is treated
as an independent single-token decode "sequence". The whole batch then runs as
one ordinary `q_len == 1` decode launch:

- **`cu_seqlens_q`** → `arange(batch * q_len + 1)` — one query per
  pseudo-sequence. The pseudo-sequence order already matches `q`'s row order
  (`q[i * q_len + j]`), so the output needs no reordering and a caller-provided
  `out` is written in place.
- **`seqused_k`** → per-position causal KV lengths. Query token `j`
  (0-indexed within the `q_len` window) attends causally to
  `seqused_k[i] - (q_len - 1 - j)` tokens, computed via broadcasting into a
  `[batch, q_len]` matrix and flattened to `[batch * q_len]`.
- **`block_table`** → `repeat_interleave(q_len, dim=0)` so each
  pseudo-sequence points at its parent sequence's blocks.

Causality is expressed entirely through the per-position `seqused_k`, so the
decode kernel is invoked with `causal=False` and produces results identical to
the causal reference.

### Correctness intuition

`seqused_k[i]` follows the standard vLLM paged convention: it is the **total**
KV length *after* the current tokens' K/V are written into the cache, i.e.
`context + q_len`. Then:

```
seqused_k[i] - (q_len - 1 - j) = (context + q_len) - (q_len - 1 - j)
                              = context + j + 1
```

which is exactly the number of KV positions `[0, context + j]` that spec token
`j` may attend to under a causal mask.

## Scope

The trigger is purely structural, so the path is **not** limited to speculative
decoding — any small, uniform-`q_len`, causal, paged batch qualifies (e.g.
fresh prefill with equal prompt lengths, or uniform-chunk chunked prefill). It
activates only when all of the following hold:

- paged KV cache (`block_table` provided) and `seqused_k` provided
- `causal=True`
- uniform query length (`q.shape[0] == batch * max_seqlen_q`)
- `1 < max_seqlen_q <= VLLM_XPU_SPEC_DECODE_MAX_QLEN` (default `16`)
- no `return_softmax_lse`, `softcap`, `alibi_slopes`, `q_v`, `q_descale`, or
  `scheduler_metadata`

Anything that does not match falls through to the existing chunk-prefill /
decode dispatch unchanged.

### Tuning knob

`VLLM_XPU_SPEC_DECODE_MAX_QLEN` (default `16`) caps the query length eligible
for the fast path. The decode path rescans the KV cache per query token (no KV
tile reuse across the Q block), so its cost grows roughly linearly in `q_len`;
beyond the measured crossover (~`q_len` 32 on Xe2) chunk-prefill is faster. Set
the variable to `0`/`1` to disable the fast path entirely.

## Performance

Measured on Intel Arc B580 (Xe2), batch = 32, single `varlen_fwd` launch,
speedup of the decode path over the chunk-prefill path:

| `q_len` | kv = 2048 | kv = 8192 |
| ------: | --------: | --------: |
|       2 |    6.93×  |    5.55×  |
|       4 |    4.66×  |    4.57×  |
|       6 |    3.93×  |    3.90×  |
|       8 |    3.51×  |    3.32×  |
|      10 |    3.04×  |    2.80×  |

The path continues to win up to ~`q_len` 32; the default threshold of 16 keeps
it comfortably inside the win region.

## Testing

- Added `test_spec_decode_uniform_qlen` (288 cases) which asserts both that the
  fast path is taken and that the output matches the causal reference across
  `q_len` 2/3/4, GQA ratios (8:2, 16:1), block sizes 16/32/64, sliding window,
  attention sink, and fp16/bf16.
- Verified single-launch behavior, in-place caller-provided `out`, and that
  non-uniform batches correctly fall through to the existing path.
- Verified correctness for pure prefill (`context = 0`) and uniform-chunk
  chunked-prefill shapes.
- No regressions: existing `test_varlen_with_paged_kv` and
  `test_decode_with_paged_kv` suites (2804 cases) still pass.
- `ruff` / `isort` / `mypy` pre-commit hooks pass.

## Files changed

- `vllm_xpu_kernels/flash_attn_interface.py` — add `_spec_decode_varlen_fwd`
  helper, the `_SPEC_DECODE_MAX_QLEN` knob, and the dispatch guard.
- `tests/flash_attn/test_flash_attn_varlen_func.py` — add
  `test_spec_decode_uniform_qlen` regression test.

### DIFF
--- a/tests/flash_attn/test_flash_attn_varlen_func.py
+++ b/tests/flash_attn/test_flash_attn_varlen_func.py
@@ -1241,3 +1241,113 @@ def test_decode_with_cross_layer_paged_kv(
     torch.testing.assert_close(output, ref_output, atol=atol, rtol=rtol), \
         f"{torch.max(torch.abs(output - ref_output))}"
     torch.xpu.empty_cache()
+
+
+# Speculative-decoding fast path: a uniform, small, causal query length per
+# sequence (q_len = num_speculative_tokens + 1) must be routed to the split-K
+# decode kernel and still match the causal reference exactly.
+@pytest.mark.parametrize("q_len", [2, 3, 4])
+@pytest.mark.parametrize("seq_lens", [[(1328, ), (18, ), (463, )],
+                                      [(1025, ), (523, ), (37, ), (200, )]])
+@pytest.mark.parametrize("num_heads", [(8, 2), (16, 1)])
+@pytest.mark.parametrize("head_size", [128])
+@pytest.mark.parametrize("block_size", [16, 32, 64])
+@pytest.mark.parametrize("window_size", [(-1, -1), (256, 0)])
+@pytest.mark.parametrize("dtype", DTYPES, ids=format_tc)
+@pytest.mark.parametrize("is_sink", [False, True])
+@pytest.mark.parametrize("num_blocks", [4096])
+@torch.inference_mode()
+def test_spec_decode_uniform_qlen(
+    q_len: int,
+    seq_lens: list[tuple[int]],
+    num_heads: tuple[int, int],
+    head_size: int,
+    block_size: int,
+    window_size: tuple[int, int],
+    dtype: torch.dtype,
+    is_sink: bool,
+    num_blocks: int,
+) -> None:
+    import vllm_xpu_kernels.flash_attn_interface as fai
+
+    torch.set_default_device("xpu")
+    torch.xpu.set_device("xpu:0")
+    torch.manual_seed(4242)
+
+    num_seqs = len(seq_lens)
+    query_lens = [q_len] * num_seqs
+    kv_lens = [x[0] for x in seq_lens]
+    num_query_heads, num_kv_heads = num_heads
+    assert num_query_heads % num_kv_heads == 0
+    max_kv_len = max(kv_lens)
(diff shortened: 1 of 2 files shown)


yma11:
Thanks and LGTM. Please handle pre-commit failure.

YizhouZ (approved):

yma11 (approved):

PR #458, opened 2026-07-13
[XPU][LoRA] Add single-launch multi-slice lora_shrink/lora_expand ops

chaojun-zhang:
## Summary
This PR adds XPU-native multi-slice LoRA operators and corresponding Punica tests. The new operators process all slices in one kernel launch instead of iterating per slice in Python.

# Motivation

Per-slice dispatch can introduce extra launch overhead and is harder to use with compile/runtime paths that are sensitive to repeated view mutation. A single-launch multi-slice implementation provides a cleaner execution model for XPU LoRA shrink/expand.

# Impact

- Introduces new XPU LoRA op entry points for multi-slice workloads
- Keeps existing per-slice BGMV ops intact (will be remove after vllm change to use signle-launch multi-slice ops)
- Expands LoRA test coverage for XPU Punica paths

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -621,6 +621,7 @@ if(XPU_SPECIFIC_KERNELS_ENABLED)
       "csrc/xpu/torch_bindings.cpp"
       "csrc/xpu/lora/lora_shrink.cpp"
       "csrc/xpu/lora/lora_expand.cpp"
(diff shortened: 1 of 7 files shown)


mayuyuace (approved):

Liangliang-Ma (approved):

jikunshang (approved):
minor: maybe we can reduce some test to avoid duplicate with vllm side.

PR #472, opened 2026-07-21
add fp8 per tensor for remap_hidden_states

mayuyuace:


jikunshang (approved):

tests/fused_moe/test_remap_hidden_states.py
```diff
@@ -123,6 +124,8 @@ def test_remap_hidden_states(num_rows, hidden_size, total_experts_num, topk,
                                    device=DEVICE,
                                    dtype=torch.uint8).view(
                                        torch.float8_e8m0fnu)
+        elif recipe == "fp8":
+            scales = None
```
Liangliang-Ma, line 128: why does scales = None here?

mayuyuace: Because one tensor share a scale.
The scale shape = [1].

Liangliang-Ma (approved):

Issue #271, opened 2026-04-13
TurboQuant KV cache quantization — feasibility report and optimization plan for Intel XPU


bryanvine:
### Summary

TurboQuant (vLLM [PR #38479](https://github.com/vllm-project/vllm/pull/38479)) is a data-oblivious KV cache quantization scheme based on Walsh-Hadamard rotation + Lloyd-Max scalar quantization, giving 3–5× KV memory compression.

We ported the upstream Triton kernels to Intel XPU (Arc Pro B70, Xe2 Battlemage) and confirmed they compile and produce correct results without any kernel modifications. Details, reproducible tests, and benchmark data are in this public repo: https://github.com/bryanvine/turboquant-xpu

Filing this issue to:
1. Document that TurboQuant works today on XPU via the upstream Triton path
2. Share benchmark data on what Intel users gain (KV capacity) and pay (throughput)
3. Propose a native SYCL optimization plan to close the throughput gap

### Functionality — works on XPU today

Using vLLM 0.19.0 + `intel-xpu-backend-for-triton` (bundled in the `intel/vllm:latest` image):

- All 3 TurboQuant Triton kernels (fused store, decode stage1, stage2 reduction) compile cleanly on Intel's SPIRV backend
- `tl.float8e4nv` works on XPU for FP8 key paths
- `tl.reshape` + grouped sum works for 3-bit MSE packing
- 2D scatter-gather from byte-addressable cache works
- Online softmax accumulation produces bit-correct results vs reference

Tested presets (all confirmed working):
- `turboquant_k8v4` (FP8 keys + 4-bit values)
- `turboquant_k3v4_nc` (3-bit MSE keys + 4-bit values + norm correction)
- `turboquant_3bit_nc` (3-bit symmetric)

#### Qwen3-30B-A3B-Instruct-2507-gptq-4bit (MoE, head_dim=128)

KV cache capacity at max-model-len=262144:

| Config | KV budget | KV tokens | vs FP16 |
|---|---:|---:|---:|
| FP16 baseline | ~10 GiB | ~65K | 1.0× |
| **TQ k3v4_nc** | 11.8 GiB | **549,888** | **~8.5×** |

Throughput (90s per concurrency level, 16-prompt mix):

| C | FP16 baseline (+EAGLE3) | TQ k3v4_nc (no spec) | ratio |
|---:|---:|---:|---:|
| 1 | ~24 tok/s | 8.7 tok/s | 0.36× |
[...]

PR #476, opened 2026-07-23
[GDN][Conv1d] Some optimization for conv1d kernel in gdn

HeJunyan:
After optimization, we get latency decrease about: Prefill −17%, mix −12%, spec −5%.


greglechin:
End-to-end measurement of this change on Intel Arc Pro B70.

**Setup.** `bmg-g31-a0` (Xe2), single card. vLLM `5a4c8d99`, vllm-xpu-kernels
`95d80c7a` plus the `csrc/` portions of **both #523 and #476** applied together
(benchmark and test files stripped; one conflicts with that pin). Model is
Qwen3.8-27B W4A16 GPTQ — Qwen3.5 hybrid, 48 GDN + 16 full-attention layers,
`head_dim` 256, `linear_key_head_dim` 128, fp8 KV, MTP with
`num_speculative_tokens=2`, XPU graph capture enabled. llama-benchy 0.4.0,
`--pp 4096 --tg 256 --concurrency 1 --runs 3 --no-cache --exact-tg`.

Both PRs were applied together, so this is their combined effect and cannot be
attributed per-PR.

**Prefill, before and after:**

| depth | baseline t/s | patched t/s | delta |
|---:|---:|---:|---:|
| 0 | 2131.58 ± 0.56 | 2015.60 ± 69.93 | −5.4% (±70) |
| 4,096 | 2079.19 ± 1.85 | 2063.38 ± 8.22 | −0.8% |
| 16,384 | 1858.93 ± 2.26 | 1854.93 ± 6.21 | −0.2% |
| 32,768 | 1630.89 ± 0.07 | 1636.22 ± 1.54 | +0.3% |
| 65,536 | 1311.72 ± 0.42 | 1313.66 ± 0.96 | +0.1% |
| 131,072 | 941.53 ± 0.23 | 942.60 ± 0.03 | +0.11% |

`t/s` is `(depth + pp) / est_ppt`, the whole-context prefill rate. Decode moved
between −6.7% and +13.5% across the same depths, all under 3 sigma.

**The patched kernels are in the binary.** Hashing the shared objects in both
images:

```
libgdn_attn_kernels_xe_2.so   e07117ca -> ec1796da   CHANGED
_xpu_C.abi3.so                66201028 -> 44be670d   CHANGED
all others                                            identical
```

**Output is byte-identical.** Six fixed prompts at temperature 0 / seed 42,
including needle-in-haystack retrieval at 7K / 27K / 94K prompt tokens, produce
identical text before and after. No rounding drift.

**Arithmetic on the null result.** If the GDN chunk kernels are X% of prefill
time and this improves them 20%, the end-to-end gain is 0.2X. An observed gain
of ≤0.3% bounds them at ≤1.5% of prefill on this model; +0.11% at 131K suggests
under 1%.

That is consistent with the work distribution at 131K context on this
architecture: approximately 67% dense linear layers, 33% full attention. The
GDN `in_proj`/`out_proj` GEMMs fall inside the linear share, leaving the chunk
kernels covering only the O(n) recurrence.

For reference, the same kernel-level versus dispatch-level gap appears in
[#271](https://github.com/vllm-project/vllm-xpu-kernels/issues/271): a fused
TurboQuant kernel measured 2.71x/4.22x standalone and 1.07x/2.04x at the vLLM
backend integration layer.

I can re-run against a different shape, or with only one of the two applied, if
that is useful. An end-to-end figure alongside the kernel-level one — on
whatever model was used for tuning — would let readers judge what to expect at
the serving layer.

PR #479, opened 2026-07-23
[ATTN] Align `softmax_lse` native layout with CUDA FlashAttention convention

chaojun-zhang:
### Background

`flash_attn_varlen_func` currently allocates and writes the `softmax_lse` tensor in layout `(total_seqlen_q, num_heads_q)`. However, downstream consumers that follow the CUDA/upstream FlashAttention layout convention (e.g., vLLM's `merge_attn_states` / `mask_empty_context`) expect layout `(num_heads_q, total_seqlen_q)`.

This mismatch has caused real crashes in the following scenario:

- MLA (DeepSeek‑V2/V3) chunked‑prefill context processing invokes `mask_empty_context`;
- when a request contains an empty context chunk, the shape mismatch triggers a `RuntimeError` inside `output.masked_fill_`, ultimately leading to severe accuracy degradation.

### Fix

Correct the layout at the Flash Attention output source, so that the native output naturally conforms to downstream expectations, avoiding the need for subsequent transposition or manual fixes.

### Concrete changes

- `csrc/flash_attn/flash_api.cpp`: allocate `softmax_lse` as `(num_heads_q, total_seqlen_q)` instead of `(total_seqlen_q, num_heads_q)`.

- `csrc/xpu/attn/xe_2/fmha_xe2.cpp`, `csrc/xpu/attn/xe_2/chunk_prefill.hpp`: `lse_stride` now takes the stride along the query row dimension (`= total_seqlen_q`) rather than along the head dimension.

- `csrc/xpu/attn/xe_2/kernel/chunk_prefill_kernel.hpp`: the SYCL epilogue writes directly to `lse[head_q][global_q]` instead of `lse[global_q][head_q]`.

- `vllm_xpu_kernels/flash_attn_interface.py` (Python fallback): natively compute and concatenate `lse` in `(heads, query_len)` directly, rather than computing in that layout and then transposing.

- `tests/flash_attn/test_flash_attn_varlen_func.py`: update the shape assertions in the reference implementation `ref_softmax_lse` and in `test_varlen_with_softmax_lse` to the new `(num_heads_q, total_seqlen_q)` convention.

### DIFF
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -195,8 +195,12 @@ std::vector<at::Tensor> mha_varlen_fwd(
   if (return_softmax) {
     int total_seqlen_q = q.size(0);
     int num_heads_q = q.size(1);
+    // (nheads, total_seqlen_q) matches the CUDA/upstream FlashAttention
+    // softmax_lse convention, and the writer kernel writes directly in
+    // this layout (see chunk_prefill_kernel.hpp), so no transpose/copy is
+    // needed on the caller side (e.g. vLLM's xpu_ops wrapper).
     softmax_lse_opt = torch::empty(
-        {total_seqlen_q, num_heads_q},
+        {num_heads_q, total_seqlen_q},
         q.options().dtype(at::kFloat).device(q.device()));
   }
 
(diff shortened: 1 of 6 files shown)


baodii (commented):
LGTM. @YizhouZ please take a look.

YizhouZ (approved):

baodii (approved):

jikunshang:
please wait a second for merging. I am checking.

jikunshang:
offline aligned. no further comments from my side. merged.

PR #490, opened 2026-07-27
[XPU] Add block FP8 scale support to grouped GEMM

kfojcik-intel:

## Purpose
This PR adds support for block-quantized FP8 scales in the XE2 grouped GEMM kernel, enabling models with block FP8 MoE layers (e.g., MiniMaxAI/MiniMax-M2.7) to use the native CUTLASS-based grouped GEMM path instead of WA of falling back to the reference fused MoE implementation.

- `grouped_gemm_xe2_interface.hpp`: Accept both 1D per-tensor [E] and 3D block
[E, K//group_size, N//group_size] FP8 scale tensors. Validate square block
quantization and compute group_size.
- `grouped_gemm_xe2.hpp`: Add block FP8 dispatch path with correct per-expert
scale offset calculation (expert_id * group_num_k * group_num_n).
- `gemm_xe2.hpp`: New xe_gemm_fp8_block_scale kernel with partial accumulator
pattern — applies block scales at 128-element boundaries during the GEMM
computation.
## Test Result
Profiled on Arc Pro B70, MiniMaxAI/MiniMax-M2.7, 256 decode tokens shows ~4× shorter XPU time and ~13× shorter CPU time vs the ref_fused_moe fallback.


csrc/xpu/grouped_gemm/xe_2/gemm_xe2.hpp
```diff
@@ -233,6 +233,179 @@ CUTE_DEVICE void xe_gemm(
...
+    class GmemTiledCopyB,
+    class GmemTiledCopyC,
+    int GroupSize,
+    class ATensor,
+    class BTensor,
+    class DTensor,
+    class TiledMMA,
+    typename ElementS,
+    typename ElementBI>
+CUTE_DEVICE void xe_gemm_fp8_block_scale(
```
mayuyuace: I do not think xe_gemm_fp8_block_scale is necessary.
And scale also need prefetch to get best performance.

mayuyuace:
Please provide performance data.

mayuyuace:
And there is already a PR for block fp8 and mxfp8:
https://github.com/vllm-project/vllm-xpu-kernels/pull/488

csrc/xpu/grouped_gemm/xe_2/gemm_xe2.hpp
kfojcik-intel: Thanks for the review @mayuyuace. Addressed both points in the latest commit:

- Merged `xe_gemm_fp8_block_scale` into` xe_gemm_4bits` — the standalone function is removed (~170 lines). Block FP8 is now selected via` if constexpr` (`is_fp8_block_scale`) which detects TB ∈ {float_e4m3_t, float_e5m2_t} && ElementS == float. The existing INT4/MXFP4 pre-multiply-B path is unchanged.

- Added scale prefetch for block FP8 — uses `make_block_2d_prefetch` on the scale tensor, same pattern as INT4/MXFP4 but adapted for the [K/group_size, N/group_size] layout.

Also added test_xe_grouped_gemm_fp8_block covering multiple shapes, expert counts, dtypes, and bias configurations.

*Edit*
I removed scale prefetch: I believe make_block_2d_prefetch has a minimum shape requirement that block FP8 scales can't satisfy with current policies — SG_N / group_size yields at most 2 (SG_N=64, group_size=32), and I think the 2D block prefetch needs ≥4 elements in the first dimension. For group_size=128/256 the condition compiles out entirely. The block scale tensor is only a few KB per expert so L2 should handle it without explicit prefetch. Let me know if I'm wrong on the HW minimum.

kfojcik-intel:
> And there is already a PR for block fp8 and mxfp8:
> #488

Thanks for the pointer. I was not aware of #488 — however, our approaches complement rather than conflict:

MXFP8 (E8M0 uint8 scales, gs=32): #488 handles this correctly via the pre-multiply-B path in xe_gemm_4bits. No overlap with this PR.

Block FP8 (float32 scales, gs=128, [K/gs, N/gs]): #488 dequantizes weights to bf16/fp16 at __init__ time and then runs W16A16 GEMM. This PR keeps weights in FP8 and uses native FP8 DPAS with a partial-accumulator + block-scale-flush approach:

- No extra memory — weights stay in FP8 (half the storage vs. init-time dequant)
- Hardware FP8 DPAS — leverages XE2 FP8 matrix engines directly
- Lower bandwidth — reads 1-byte weights instead of 2-byte

The two PRs can coexist: #488 for MXFP8, this PR for block FP8 native compute. How would you like to see the merge strategy here? Should we:

1. Merge #488 first (MXFP8) and rebase this PR on top?
2. Combine into a single PR covering both MXFP8 and block FP8?

Happy to adapt to whatever workflow works best for you.

PR #493, opened 2026-07-27
refactor fused_moe benchmark

mayuyuace:
1. Update xpu_fused_moe_CalKernelTime.
2. Use triton.testing.do_bench and statistics.median instead of event to get stable perf data.

### DIFF
--- a/benchmark/benchmark_cutlass_fused_moe.py
+++ b/benchmark/benchmark_cutlass_fused_moe.py
@@ -4,6 +4,7 @@
 
 # isort: off
 import gc
+import statistics
 
 import torch
 import triton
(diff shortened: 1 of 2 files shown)


jikunshang:
cc @1pikachu

1pikachu:
Do we need to switch all the other kernels to triton.do_bench as well?  @jikunshang

mayuyuace:
@1pikachu 
I think triton.do_bench is better, it can clear up L2 cache automatically and get median time for stable perf data.
And you do not need set iteration, because triton.do_bench can get it automatically.

jikunshang (approved):

Liangliang-Ma (approved):

PR #494, opened 2026-07-27
fused moe gate up

mayuyuace:
Add kernel fused_moe_up to fuse moe gemm1+clamp+act_fun.


csrc/xpu/torch_bindings.cpp
```diff
@@ -278,6 +278,13 @@ TORCH_LIBRARY_EXPAND(TORCH_EXTENSION_NAME, xpu_ops) {
       "deepseek_fused_indexer_q_rope_mxfp4",
       torch::kXPU,
       &deepseek_fused_indexer_q_rope_mxfp4);
+
+  xpu_ops.def(
+      "fused_moe_up(Tensor ptr_A, Tensor? ptr_A_scale, "
```
jikunshang: `fused_moe_gate_up`

mayuyuace: Changed.
And I am asking validation to test MOE models performance with this PR.

PR #499, opened 2026-07-29
Fix getMemoryInfo to work with older PVC drivers

frost-intel:
## Purpose
`csrc/utils/mem_info.cpp` unconditionally used  `ze_device_usablemem_size_ext_properties_t`, which only exists in newer Level Zero headers. On stacks that predate the device-usablemem-size extension the `_C` extension fails to compile outright.

This PR instead falls back to the torch `c10::xpu::XPUCachingAllocator::get()->getMemoryInfo()` if usablemem is not supported.

This is required for Max 1550 GPUs  (PVC) since L0 is only updated for bug fixes, no version upgrades.

~~I used a cmake compile test to define `VLLM_XPU_HAS_ZE_USABLEMEM`. If you'd prefer gating by a specific level-zero version, let me know.~~

UPDATE: New approach is to define the missing types in a way that only changes behavior at runtime, rather than build.

### DIFF
--- a/csrc/utils/mem_info.cpp
+++ b/csrc/utils/mem_info.cpp
@@ -5,19 +5,6 @@
 
 #include <iostream>
 
-// Level Zero headers predating the device-usablemem-size-properties extension
-// declare neither of these, which is a compile error rather than something a
-// runtime check can guard, so spell out the spec-fixed stype and layout here.
-namespace {
-constexpr auto kUsableMemStype = static_cast<ze_structure_type_t>(0x00020041);
-
-struct UsableMemProps {
-  ze_structure_type_t stype;
-  void* pNext;
-  uint64_t currUsableMemSize;
-};
-}  // namespace
-
 size_t getTotalMemory(ze_device_handle_t& device) {
   uint32_t memoryCount = 0;
   zeDeviceGetMemoryProperties(device, &memoryCount, nullptr);
@@ -36,21 +23,22 @@ size_t getTotalMemory(ze_device_handle_t& device) {
   return totalMemory;
 }
 
-// Zero means the extension is unavailable: a driver that does not implement it
-// ignores the unrecognized pNext and still reports success, leaving the
-// zero-initialized field untouched.
+// A driver that does not implement the extension leaves the zero-initialized
+// size untouched and still returns success, so zero means unavailable.
 size_t getUsableMemory(ze_device_handle_t& device) {
+#ifdef ZE_DEVICE_USABLEMEM_SIZE_PROPERTIES_EXT_NAME
   ze_device_properties_t deviceProperties{};
-  UsableMemProps usableMemProps{};
+  ze_device_usablemem_size_ext_properties_t usableMemProps{};
 
-  usableMemProps.stype = kUsableMemStype;
+  usableMemProps.stype = ZE_STRUCTURE_TYPE_DEVICE_USABLEMEM_SIZE_EXT_PROPERTIES;
   deviceProperties.stype = ZE_STRUCTURE_TYPE_DEVICE_PROPERTIES;
   deviceProperties.pNext = &usableMemProps;
 
-  if (zeDeviceGetProperties(device, &deviceProperties) != ZE_RESULT_SUCCESS) {
-    return 0;
-  }
+  zeDeviceGetProperties(device, &deviceProperties);
   return usableMemProps.currUsableMemSize;
+#else
+  return 0;
+#endif
 }
 
 std::tuple<int64_t, int64_t> getMemoryInfo(int64_t device_index) {
--- a/csrc/utils/mem_info.cpp
+++ b/csrc/utils/mem_info.cpp
@@ -4,7 +4,8 @@
 #include <sycl/sycl.hpp>
 
 #include <iostream>
-
+#include <limits>
+#include <tuple>
 #ifdef VLLM_XPU_HAS_ZE_USABLEMEM
 size_t getTotalMemory(ze_device_handle_t& device) {
   uint32_t memoryCount = 0;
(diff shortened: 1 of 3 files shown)


mayuyuace:
@xinyu-intel @jikunshang 
Do you have any comments on this PR for PVC?

baodii (approved):
LGTM

xinyu-intel (approved):

xinyu-intel (changes requested):
is it possible to check runtime instead of build time?

jikunshang:
> is it possible to check runtime instead of build time?

agree. can you try below code to get ze version and do some runtime dispatch?

```
#include <level_zero/ze_api.h>
#include <iostream>
#include <vector>

int main() {
    // 1. Initialize the Level Zero driver
    ze_result_t result = zeInit(ZE_INIT_FLAG_GPU_ONLY);
    if (result != ZE_RESULT_SUCCESS) return -1;

    // 2. Discover and query all available driver handles
    uint32_t driverCount = 0;
    zeDriverGet(&driverCount, nullptr);
    if (driverCount == 0) return -1;

    std::vector<ze_driver_handle_t> drivers(driverCount);
    zeDriverGet(&driverCount, drivers.data());

    // 3. Retrieve and print the supported API version of the primary driver
    ze_api_version_t apiVersion;
    result = zeDriverGetApiVersion(drivers[0], &apiVersion);
    
    if (result == ZE_RESULT_SUCCESS) {
        // Extract Major and Minor numbers using specification-defined macros
        uint32_t major = ZE_MAJOR_VERSION(apiVersion);
        uint32_t minor = ZE_MINOR_VERSION(apiVersion);
        std::cout << "Level Zero API Version: " << major << "." << minor << std::endl;
    }

    return 0;
}

```

frost-intel:
@jikunshang  @xinyu-intel See my update. I can rework this to be a runtime check, though it requires a typdef since there are missing symbols. Then we still fall back to `XPUCachingAllocator::getMemoryInfo`

xinyu-intel:
> @jikunshang @xinyu-intel See my update. I can rework this to be a runtime check, though it requires a typdef since there are missing symbols. Then we still fall back to `XPUCachingAllocator::getMemoryInfo`

I think the fallback logic in https://github.com/intel/llvm/pull/21339/changes can help.

frost-intel:
@xinyu-intel @jikunshang — this does check at runtime now; it just uses the returned value as the signal rather than a version query. Two cases:

**1. New build headers, old driver at runtime** (i.e. downloading public wheel). `zeDeviceGetProperties` will returns `ZE_RESULT_SUCCESS` and leave `currUsableMemSize = 0`:

```cpp
// compute-runtime: level_zero/core/source/device/device.cpp
} else if (extendedProperties->stype == ZE_STRUCTURE_TYPE_DEVICE_USABLEMEM_SIZE_EXT_PROPERTIES) {
    usableMemProperties->currUsableMemSize = this->neoDevice->getUsableMemorySize();
}
```

**2. Old build headers.** Since some types are not defined, we gate as follows:

```cpp
size_t getUsableMemory(ze_device_handle_t& device) {
#ifdef ZE_DEVICE_USABLEMEM_SIZE_PROPERTIES_EXT_NAME
  ...query...
#else
  return 0;
#endif
}
```

Both cases converge on `free == 0`, so one fallback covers them:

```cpp
std::tuple<int64_t, int64_t> getMemoryInfo(int64_t device_index) {
  ...
  size_t free = getUsableMemory(level_zero_device);
  size_t total = 0;
  if (free == 0) {
    std::tie(free, total) = c10::xpu::XPUCachingAllocator::get()->getMemoryInfo(...);
  }
```

Regarding https://github.com/intel/llvm/pull/21339/changes:  There are multiple things happening in that code, so I'm not sure if you were referring to the `#ifdef` gate (which I used), or checking the extension properties (which I did not use, since it's redundant with the current code). I'm happy to add the extension check if you'd prefer for symmetry with the existing L0 code.

I also added an unconditional `assert free > 0` to the test as a basic sanity check for PVC, since both branches are BMG-only. This test wouldn't prevent this kind of error without changing the PVC CI image to the LST2-AI image, which is something I'm looking into, watch for a future PR.

csrc/utils/mem_info.cpp
```diff
@@ -32,15 +36,26 @@ size_t getUsableMemory(ze_device_handle_t& device) {
...
   const auto& device =
       c10::xpu::get_raw_device(static_cast<c10::DeviceIndex>(device_index));
   auto level_zero_device =
       sycl::get_native<sycl::backend::ext_oneapi_level_zero>(device);
-  const size_t free = getUsableMemory(level_zero_device);
-  const size_t total = getTotalMemory(level_zero_device);
+
+  size_t free = getUsableMemory(level_zero_device);
+  size_t total = 0;
+  if (free == 0) {
```
mayuyuace: Free of l0 on BMG does not equal the free value of torch.
This may raise confusing behavior.

frost-intel: Agreed. Updated code now checks for extension availability. It's a bit more complicated, but safer overall.

tests/test_get_memory_info.py
```diff
@@ -12,6 +12,11 @@
 def test_get_memory_info(device) -> None:
     free, total = torch.ops._C_cache_ops.getMemoryInfo(device)
 
+    # Drivers without the usable-memory extension report 0; the fallback must
+    # still produce real numbers on every device.
+    assert free > 0
```
mayuyuace, line 17: Is it possible to provide precise values ​​for PVC to ensure the correct functionality of the UT?

mayuyuace:
LGTM.
@jikunshang @xinyu-intel 
Do you have any further comments?

xinyu-intel (approved):

mayuyuace (approved):

PR #507, opened 2026-08-05
remove some benchmark configs which don't run on B60

mayuyuace:


Liangliang-Ma (commented):
why we not just delete them but comment out?

mayuyuace:
> why we not just delete them but comment out?

In case CRI need these configs.
Deleting is also fine to me.
What do you prefer?

Liangliang-Ma:
> > why we not just delete them but comment out?
> 
> In case CRI need these configs. Deleting is also fine to me. What do you prefer?

we can delete them. Post PO model owner will add if needed.

mayuyuace:
> > > why we not just delete them but comment out?
> > 
> > 
> > In case CRI need these configs. Deleting is also fine to me. What do you prefer?
> 
> we can delete them. Post PO model owner will add if needed.

Fine. Useless configs have been removed.

zufangzhu (approved):

Liangliang-Ma (approved):

PR #508, opened 2026-08-06
Check if FP8 block GEMM [k,n] are divisible by block scales

frost-intel:
## Purpose
Adds a safety check for fp8 Block GEMM kernels to ensure that the block scales give evenly divide `[k, n]`. 

**NOTE:** This check does not guarantee accuracy if the block scale **does** evenly divide `[k, n]`, since if weights are not properly padded then a silent accuracy error can still occur. See https://github.com/vllm-project/vllm/pull/49596 for a fix and more details. A check to guarantee accuracy could be achieved by passing the weight scale sizes to every function using block quant, but since fixing this mistake would need to happen in the calling code in vLLM, not in vllm-xpu-kernels, and this would require a major API change for many functions, I didn't implement this.

## Test Plan
Added fp8_gemm test to check for error raised on indivisible inputs.

## Test Result
```
collected 1277 items / 1268 deselected / 9 selected                                                                          

tests/test_fp8_gemm_onednn.py::test_fp8_block_indivisible_scale_raises[n_indivisible-w8a8-dtype0-fp8_dtype0] PASSED    [ 11%]
tests/test_fp8_gemm_onednn.py::test_fp8_block_indivisible_scale_raises[n_indivisible-w8a16-dtype0-fp8_dtype0] PASSED   [ 22%]
tests/test_fp8_gemm_onednn.py::test_fp8_block_indivisible_scale_raises[n_indivisible-bmm-dtype0-fp8_dtype0] PASSED     [ 33%]
tests/test_fp8_gemm_onednn.py::test_fp8_block_indivisible_scale_raises[k_indivisible-w8a8-dtype0-fp8_dtype0] PASSED    [ 44%]
tests/test_fp8_gemm_onednn.py::test_fp8_block_indivisible_scale_raises[k_indivisible-w8a16-dtype0-fp8_dtype0] PASSED   [ 55%]
tests/test_fp8_gemm_onednn.py::test_fp8_block_indivisible_scale_raises[k_indivisible-bmm-dtype0-fp8_dtype0] PASSED     [ 66%]
tests/test_fp8_gemm_onednn.py::test_fp8_block_indivisible_scale_raises[kn_indivisible-w8a8-dtype0-fp8_dtype0] PASSED   [ 77%]
tests/test_fp8_gemm_onednn.py::test_fp8_block_indivisible_scale_raises[kn_indivisible-w8a16-dtype0-fp8_dtype0] PASSED  [ 88%]
tests/test_fp8_gemm_onednn.py::test_fp8_block_indivisible_scale_raises[kn_indivisible-bmm-dtype0-fp8_dtype0] PASSED    [100%]

============================================= 9 passed, 1268 deselected in 0.83s =============================================
```


jikunshang:
cc @zufangzhu

tests/test_fp8_gemm_onednn.py
```diff
@@ -483,3 +483,49 @@ def test_fp8_gemm_w8a16_block(fp8_dtype, dtype, is_nt, batch, group_size,
     # bf16 has lower mantissa precision (7-bit vs 10-bit for f16), which causes
     # larger accumulated error in the bf16+fp8 block matmul kernel.
     torch.testing.assert_close(output_fp8, output_ref, atol=6e-2, rtol=6e-2)
+
+
+@pytest.mark.parametrize("fp8_dtype", [torch.float8_e4m3fn])
+@pytest.mark.parametrize("dtype", [torch.bfloat16])
+@pytest.mark.parametrize("mode", ["w8a8", "w8a16", "bmm"])
+@pytest.mark.parametrize("m, n, k, scale_gk, scale_gn", (
```
Yejing-Lai: Can you distribute these testcases among the existing UT instead of creating a new one?

frost-intel: Do you mean to fold these test cases into existing block tests? If so, I'd push back on that idea. The existing block tests build their scale grid with `fp8_block_quant_2d()` so an indivisible grid can't be produced through that path. I'd need to construct the scale tensor and skip the dequant reference, as well as adding early-exit branches inside other tests that are asserting numeric accuracy.

I don't know that the tests are the most impactful, if you think they're unnecessary I could remove.

However, consider that these tests aren't actually computing any workload, so they execute very quickly (9 tests in < 0.5 s on Max 1550).

Yejing-Lai: Remove it is ok for me. We should not add a new UT for the torch check.

Yejing-Lai (approved):

zufangzhu (approved):

jikunshang (approved):

PR #516, opened 2026-08-11
fix(mem_cpy): use get_pointer_device to infer XPU device, fixing D2H ordering

libinta:
On Intel GPUs all XPU sub-devices share a single Level-Zero/SYCL context, so sycl::get_pointer_type(ptr, ctx) reports usm::alloc::device for a given device pointer against every device's context.  The previous first-match loop in infer_xpu_device_from_ptr() therefore always returned device 0 regardless of which device actually owns the allocation.

The resulting DeviceGuard(0) flipped the current device to 0 and vllmGetQueue() returned device 0's in-order queue -- a queue with no ordering relationship to the compute/transfer stream on the true device.  For the D2H KV-cache offload store this silently defeated the stream.wait_stream(compute) fence in gpu_worker.py, racing the D2H copy against the still-in-flight cache write and producing nondeterministic KV-restore corruption.  Under tensor parallelism ranks > 0 were affected deterministically (their cache shards always mis-inferred to device 0), which explains why the corruption was reproducible per-prompt across runs.

Fix: replace the get_pointer_type first-match loop with sycl::get_pointer_device(), which returns the actual owning device, then map the returned sycl::device back to the c10 device index.  The copy now lands on the correct device's in-order queue that wait_stream already fenced.

The fix is async and introduces no host stall or performance regression.

Verified:
- Isolated probe (qprobe3): buggy 388/400 CORRUPT -> fixed 0/400 CLEAN
- Real swap_blocks_batch under production wait_stream race: 399/400 -> 0/400
- In-tree coherent_parity self-consistency (W4A16 TP2 native offload): R1 vs R2 1/16 token-exact -> 16/16 token-exact after restore
- Full test_cache.py + test_xpu_memcpy_sync.py suites: all pass

Adds test_swap_blocks_batch_d2h_stream_ordering: a regression test that reproduces the race (concurrent slow writer + D2H on a non-zero device) and fails on the old code / passes on the fixed code.  Skipped when device_count() < 2 since the bug is invisible on single-device systems (device 0 is inferred correctly by accident).

### DIFF
--- a/csrc/utils/mem_cpy.cpp
+++ b/csrc/utils/mem_cpy.cpp
@@ -151,16 +151,27 @@ void xpuAsyncMemcpy(
   }
 }
 
-// Infer which XPU device a USM device pointer was allocated on by probing
-// each device's SYCL context.  Returns the device index on success.
-// This is O(num_xpu_devices) but avoids threading an explicit device argument
-// through the entire call chain when all callers already have the pointer.
+// Infer which XPU device a USM device pointer was allocated on.
+// Returns the device index on success.
+//
+// NOTE: on Intel GPUs all XPU (sub-)devices share a single SYCL/Level-Zero
+// context, so sycl::get_pointer_type() reports usm::alloc::device for a given
+// device pointer against EVERY device's context.  A first-match loop over
+// get_pointer_type() therefore always returns device 0, no matter which device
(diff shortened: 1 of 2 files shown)


chaojun-zhang (approved):
LGTM

jikunshang (approved):

mayuyuace (approved):

PR #523, opened 2026-08-13
[XE2][GDN] Some optimization for gdr prefill kernel

HeJunyan:
After optimization, we get latency decrease about: Prefill −20% in mean and best  -47%.


greglechin:
edit - disregard

HeJunyan:
@greglechin，so in this Qwen3.8-27B model, the dense linear layers and full attention take more than 99% time, and so this GDN performance improvement is ignorable, right ?

greglechin:
@HeJunyan yes sorry for my confusion

PR #532, opened 2026-08-17
[Qwen3.5] feat(layernorm): add fused XPU GemmaRMSNorm kernels

ccrhx4:
# [Kernel] Add fused XPU GemmaRMSNorm kernels  for the eager path

## Summary
Adds fused XPU ops for GemmaRMSNorm so the enforce-eager path gets a single fused kernel instead of falling back to separated element-wise ops — closing the enforce-eager vs  torch.compile  prefill gap for  GemmaRMSNorm  models (e.g. Qwen3.5).

Two new ops:  gemma_rms_norm(...)  and  fused_add_gemma_rms_norm(...) . The only functional difference from the existing kernels is the Gemma unit offset: weight applied as  (1 + weight)  instead of  weight .

out = round(x * inv_rms(x) * (weight + weight_bias))

• Generic ops:  weight_bias = 0.0f  (default) — byte-identical to master.
• Gemma ops:  weight_bias = 1.0f  — Gemma's  (1 + weight) .

## Performance (enforce-eager prefill)

| Model            | TP | eager (native fallback) | eager (fused) | torch.compile |
|------------------|----|-------------------------|---------------|---------------|
| Qwen3.5-9B       | 1  | 1.256 s                 | **1.023 s**   | 1.013 s       |
| Qwen3.5-35B-A3B  | 4  | 0.580 s                 | **0.419 s**   | 0.421 s       |

## Accuracy (GSM8K, lm-eval-harness, max_gen_toks=4096)

| Model            | TP | exact_match (strict) | exact_match (flexible) |
|------------------|----|----------------------|------------------------|
| Qwen3.5-9B       | 1  | 0.8923               | 0.8901                 |
| Qwen3.6-27B      | 4  | 0.6748               | 0.6755                 |
| Qwen3.5-35B-A3B  | 4  | 0.9462               | 0.9538                 |

## Testing

•  tests/test_layernorm.py : 2304 passed (704 Gemma cases), on current  main  (post-#534). pre-commit clean.

### DIFF
--- a/tests/ops/layernorm_op.py
+++ b/tests/ops/layernorm_op.py
@@ -42,6 +42,109 @@ def dispatch_cuda_rmsnorm_func(add_residual: bool):
     return rms_norm
 
 
+def gemma_rms_norm(x: torch.Tensor, weight: torch.Tensor,
+                   variance_epsilon: float) -> torch.Tensor:
+    import tests.register_ops as ops
+    out = torch.empty_like(x)
+    ops.gemma_rms_norm(
+        out,
+        x,
+        weight,
+        variance_epsilon,
+    )
+    return out
+
+
+def fused_add_gemma_rms_norm(
+        x: torch.Tensor, residual: torch.Tensor, weight: torch.Tensor,
+        variance_epsilon: float) -> tuple[torch.Tensor, torch.Tensor]:
+    import tests.register_ops as ops
+    ops.fused_add_gemma_rms_norm(
+        x,
+        residual,
+        weight,
+        variance_epsilon,
+    )
+    return x, residual
+
+
+def dispatch_cuda_gemma_rmsnorm_func(add_residual: bool):
+    if add_residual:
+        return fused_add_gemma_rms_norm
+    return gemma_rms_norm
+
+
+class GemmaRMSNorm(CustomOp):
+    """RMS normalization for Gemma.
+
+    Two differences from RMSNorm:
+        1. x * (1 + w) instead of x * w.
+        2. (x * (1 + w)) is computed in fp32 then downcast.
+    """
+
+    def __init__(
+        self,
+        hidden_size: int,
+        eps: float = 1e-6,
+        dtype: Optional[torch.dtype] = None,
+    ) -> None:
+        super().__init__()
+        self.hidden_size = hidden_size
+        self.variance_epsilon = eps
+        if dtype is not None:
+            self.weight = nn.Parameter(torch.zeros(hidden_size, dtype=dtype))
+        else:
+            self.weight = nn.Parameter(torch.zeros(hidden_size))
+
+    @staticmethod
+    def forward_static(
+        weight: torch.Tensor,
+        variance_epsilon: float,
+        x: torch.Tensor,
+        residual: Optional[torch.Tensor],
+    ) -> Union[torch.Tensor, tuple[torch.Tensor, torch.Tensor]]:
+        orig_dtype = x.dtype
+        if residual is not None:
+            x = x + residual.to(torch.float32)
+            residual = x.to(orig_dtype)
+
+        x = x.to(torch.float32)
+        variance = x.pow(2).mean(dim=-1, keepdim=True)
+        x = x * torch.rsqrt(variance + variance_epsilon)
+        # Llama does x.to(float16) * w whilst Gemma is (x * w).to(float16)
+        # See https://github.com/huggingface/transformers/pull/29402
+        x = x * (1.0 + weight.float())
+        x = x.to(orig_dtype)
+        return x if residual is None else (x, residual)
+
+    def forward_native(
+        self,
+        x: torch.Tensor,
+        residual: Optional[torch.Tensor] = None,
+    ) -> Union[torch.Tensor, tuple[torch.Tensor, torch.Tensor]]:
+        return self.forward_static(self.weight.data, self.variance_epsilon, x,
+                                   residual)
+
+    def forward_cuda(
+        self,
+        x: torch.Tensor,
+        residual: Optional[torch.Tensor] = None,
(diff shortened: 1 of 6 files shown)


ccrhx4:
@xinyu-intel @jikunshang @wuxun-zhang Please help to review. Thanks.

jikunshang (commented):
I prefer to add a standalone gemma template method instead of reusing existing method, this make code messy.

xinyu-intel:
> I prefer to add a standalone gemma template method instead of reusing existing method, this make code messy.

agree. BTW, even if re-using, IS_GEMMA is not a clear semantic and will be break easily if any modeling changes in the future.

ccrhx4:
The gemma style rms_norm is very similar to the origin rms_norm, the only change is to add 1.0 to the learned weight parameters. There will be lots of redundant code if we want a new API. 

I suggest to rename **IS_GEMMA** to **weight_bias**, similar to flashinfer library https://github.com/flashinfer-ai/flashinfer/commit/5bf36ced5e1c33e68099f4f8380dfdeadd63c78d

xinyu-intel:
> The gemma style rms_norm is very similar to the origin rms_norm, the only change is to add 1.0 to the learned weight parameters. There will be lots of redundant code if we want a new API.
> 
> I suggest to rename **IS_GEMMA** to **weight_bias**, similar to flashinfer library [flashinfer-ai/flashinfer@5bf36ce](https://github.com/flashinfer-ai/flashinfer/commit/5bf36ced5e1c33e68099f4f8380dfdeadd63c78d)

then, it will be much better. @jikunshang what's your opinion?

ccrhx4:
And I found a divergence in rms_norm implementation on XPU and CUDA, [VLLMZ-2291](https://jira.devtools.intel.com/browse/VLLMZ-2291)

After this bug is fixed, the code change will be very small.  Just like code below:
```diff
- input_vec[j] = x_vec[j] * rms_rcp * float(weight_vec[j]);
+ input_vec[j] = x_vec[j] * rms_rcp * (weight_bias + float(weight_vec[j]));
```

jikunshang:
> > The gemma style rms_norm is very similar to the origin rms_norm, the only change is to add 1.0 to the learned weight parameters. There will be lots of redundant code if we want a new API.
> > I suggest to rename **IS_GEMMA** to **weight_bias**, similar to flashinfer library [flashinfer-ai/flashinfer@5bf36ce](https://github.com/flashinfer-ai/flashinfer/commit/5bf36ced5e1c33e68099f4f8380dfdeadd63c78d)
> 
> then, it will be much better. @jikunshang what's your opinion?

sounds good to me, thanks.

jikunshang (commented):
cc @zufangzhu

tests/ops/layernorm_op.py
```diff
@@ -42,6 +42,109 @@ def dispatch_cuda_rmsnorm_func(add_residual: bool):
...
+
+    def forward_native(
+        self,
+        x: torch.Tensor,
+        residual: Optional[torch.Tensor] = None,
+    ) -> Union[torch.Tensor, tuple[torch.Tensor, torch.Tensor]]:
+        return self.forward_static(self.weight.data, self.variance_epsilon, x,
+                                   residual)
+
+    def forward_cuda(
```
zufangzhu, line 128: Why we have this in vllm-xpu-kernels

ccrhx4: This new test class is written to match the existing  RMSNorm reference class (same file): forward_xpu  delegates to forward_cuda , which dispatches to  torch.ops._C.rms_norm.  GemmaRMSNorm  simply follows that same pattern.  Would you prefer to inline the dispatch directly in GemmaRMSNorm.forward_xpu ? And also do the same for RMSNorm?

tests/register_ops.py
```diff
@@ -23,6 +23,17 @@ def fused_add_rms_norm(input: torch.Tensor, residual: torch.Tensor,
     torch.ops._C.fused_add_rms_norm(input, residual, weight, epsilon)
 
 
+def gemma_rms_norm(out: torch.Tensor, input: torch.Tensor,
+                   weight: torch.Tensor, epsilon: float) -> None:
+    input_contiguous = input.contiguous()
+    torch.ops._C.gemma_rms_norm(out, input_contiguous, weight, epsilon)
+
```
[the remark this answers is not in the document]
ccrhx4: Fixed in 155a4ea69025afcac543319b5764e40eae0701f3

mayuyuace:
I believe the improvement in model performance is primarily due to the use of a fused kernel instead of the native implementation.
As this PR https://github.com/vllm-project/vllm/pull/53203 said, the vllm_c kernel's supports_args check (rms_no_var_size) requires weight.dtype == x.dtype, so with float32 weight and bfloat16 input the check fails and dispatch always falls back to the native PyTorch decomposition (~8 kernel launches per norm).
The difference of gemma_rms between rms is only the bias 1.

ccrhx4:
@mayuyuace You are right. After this PR is commit, I will change the vLLM side to call this gemma rms norm, instead of the native path.

jikunshang (approved):

ccrhx4:
I have created the PR in vLLM to call the new gemma rms API https://github.com/vllm-project/vllm/pull/53678

I am aware Qiming created a PR to fix the same issue in vLLM for XPU, https://github.com/vllm-project/vllm/pull/53203, but I afraid this PR changes the numeric in the native path. I have raised my concern in the comments.

Please take a look and kindly give u comments. Thanks. @xinyu-intel @jikunshang @zufangzhu @mayuyuace

jikunshang:
add to 0.1.14 https://github.com/vllm-project/vllm-xpu-kernels/issues/546

xinyu-intel (approved):

Issue #510, opened 2026-08-07
[Bug] GDN causal_conv1d rejects mixed spec/non-spec batches — MTP unusable under concurrency


SergiioB:
## Summary

The fused XPU SYCL `gdn_attention` kernel refuses any batch that mixes speculative-decode tokens with non-spec (prefill or plain decode) tokens:

```
RuntimeError: causal_conv1d does not support spec-decode and non-spec
(prefill + decode) tokens in the same invocation; the spec path and the
non-spec path are mutually exclusive
```

The check is still present on current main (`de8eae1`, `csrc/xpu/gdn_attn/gdn_attn_interface.cpp:62-65`). This makes MTP speculative decoding unusable under concurrency on hybrid-GDN models — the engine dies on the first mixed step.

## Environment

- Hardware: Intel Arc Pro B70 32 GB (Battlemage, Xe2), single card
- Container: `intel/vllm:0.21.0-xpu-int4moe`, vLLM `v0.21.1.dev18+gec426505f`
- Model: Qwen3.6-35B-A3B MoE, GPTQ-Int4 checkpoint with preserved MTP weights (30/40 layers are GDN linear attention, GQA ratio 3: num_k_heads=16, num_v_heads=48, head dim 128)
- Config: `--speculative-config '{"method":"mtp","num_speculative_tokens":1}'`, `--max-num-seqs 16`, `--max-num-batched-tokens 8192`, cudagraph PIECEWISE capture sizes 1/2/4/8

## Reproduce

1. Serve the model with MTP spec decode and `--max-num-seqs > 1`.
2. Send ≥2 concurrent chat-completion requests.
3. EngineCore dies as soon as a scheduled step mixes token classes — e.g. one request still prefilling (or on its first decode step, before draft tokens exist) while the others are spec-decoding.

Single-stream serving (`--max-num-seqs 1`) never mixes classes and is unaffected — which is why the crash only appears under load.

## Impact

MTP adds +69% single-stream decode on this model (1.69x vs no-spec baseline, 204.6 t/s tg32 on the LocalMaxxing leaderboard), but this check forces a choice between spec decode and concurrency. Effectively `--max-num-seqs 1`, which forfeits the whole point of vLLM on this card.

## Workaround attempted (and why it is not enough)

[...]

PR #402, opened 2026-06-08
[GDN] Split gdn attention into  conv1d and gated_delta_rule

HeJunyan:
The current gdn_attention includes everything and is too huge. It is inconvenient for profiling. So we split it into conv1d and gated_delta_rule. We also add test cases and benchmark for conv1d and gated_delta_rule separately. For the legacy usage, we still keep the original gdn_attention operator, which is a simple wrap of calling conv1d and gated_delta_rule.

PR #537, opened 2026-08-18
[GDN][MTP] fix split PR bug 

mayuyuace:
**Issue**: https://github.com/vllm-project/vllm-xpu-kernels/issues/510
**Guilty PR**:  https://github.com/vllm-project/vllm-xpu-kernels/pull/402
**Regression**: RuntimeError: causal_conv1d does not support spec-decode and non-spec
(prefill + decode) tokens in the same invocation; the spec path and the
non-spec path are mutually exclusive
**Fix**: add new split op and refactor logic. And add mixed UT.

**Test**:
**CMD**:
```
VLLM_ALLOW_LONG_MAX_MODEL_LEN=1\
 vllm serve Qwen/Qwen3.6-27B \
  --host 0.0.0.0 \
  --port 30088 \
  --tensor-parallel-size 2 \
  --quantization fp8 \
  --dtype float16 \
  --enforce-eager \
  --enable-auto-tool-choice \
  --reasoning-parser qwen3 \
  --tool-call-parser qwen3_xml \
  --gpu-memory-utilization 0.95 \
  --max-num-batched-tokens 8192 \
  --max-model-len 8192 \
  --max-num-seqs 4 \
  --block-size 64 \
  --mamba-ssm-cache-dtype float16 \
  --trust-remote-code \
  --enable-prefix-caching \
  --speculative-config '{"method": "dflash", "model": "z-lab/Qwen3.6-27B-DFlash", "num_speculative_tokens": 1}'

lm_eval --model local-completions --tasks gsm8k --model_args "model=Qwen/Qwen3.6-27B,base_url=http://localhost:30088/v1/completions,max_gen_toks=1024,num_concurrent=64,timeout=9999999" --limit 100

```

**Results**:
**Sever**:
image

**Client**:
image


Wei-Lin-Intel (approved):
LGTM

jikunshang (approved):

YizhouZ (approved):

PR #540, opened 2026-08-19
[LoRA][XPU] Optimize active-rank linear kernels

liu-shaojun:
## Summary

- Compute only the active adapter rank while preserving the padded stored-rank strides used by shared LoRA slots.
- Add a fused `lora_linear` entry point that combines multi-slice shrink and expand dispatch.
- Tune shrink/expand workgroup sizing and retain a per-slice fallback for large balanced rank-128 three-slice outputs.
- Add FP16/BF16 correctness coverage for one, two, and four slices, base rows, overwrite/add modes, and padded ranks.

## Motivation

A LoRA server can host rank-64 and rank-128 adapters in the same max-rank-128 storage. The existing path computes the padded rank for every request and submits shrink/expand work separately for packed QKV, MLP, and GDN slices. This makes the lower-rank adapter nearly as expensive as rank 128 and exposes substantial launch overhead during decode.

The new API separates active rank from stored row stride. It therefore trims arithmetic without changing the shared weight layout or adapter lifecycle.

## Compatibility

The new `lora_linear` op uses int32 token mappings and is capability-detected by the companion vLLM change. Deployments with an older kernel package continue to use the existing `bgmv_linear` path and int64 mappings.

## Duplicate-work check

- #458 already supplied the multi-slice foundation on `main`; this change is incremental and builds on that implementation.
- I checked the open kernel PRs and found no existing active-rank/stored-stride or fused `lora_linear` change.

## Validation

```text
XPU_KERNEL_PYTEST_PROFILER=MINI PYTHONPATH=.:/opt/venv/lib/python3.12/site-packages \
  /opt/venv/bin/python -m pytest -q tests/test_lora_ops.py
19 passed

pre-commit run --files csrc/xpu/lora/lora_ops.cpp csrc/xpu/lora/lora_ops.h \
  csrc/xpu/torch_bindings.cpp tests/lora/xpu_ops.py tests/test_lora_ops.py
all hooks passed
```

The exact source was rebuilt and tested on XPU with torch 2.12 and oneAPI 2025.3.

Combined with the companion vLLM routing change, Qwen3.6-27B TP2 eager serving produced identical generated text and the following single-request decode throughput:

| Quantization | Request | Before | After | Change |
|---|---|---:|---:|---:|
| FP8 | base | 67.31 tok/s | 102.84 tok/s | +52.8% |
| FP8 | rank-128 SLO | 51.95 tok/s | 82.23 tok/s | +58.3% |
| FP8 | rank-64 Fable | 53.12 tok/s | 88.34 tok/s | +66.3% |
| sym_int4 | base | 59.92 tok/s | 93.78 tok/s | +56.5% |
| sym_int4 | rank-128 SLO | 49.01 tok/s | 75.08 tok/s | +53.2% |
| sym_int4 | rank-64 Fable | 50.02 tok/s | 81.06 tok/s | +62.1% |

## AI assistance

GitHub Copilot assisted with implementation, test execution, benchmarking, and code review. The submitter is responsible for the final change.

## Companion change

The coordinated vLLM routing and serving change is intel-innersource/applications.ai.gpu.llm-scaler-vllm#22.

### DIFF
--- a/csrc/xpu/lora/lora_ops.cpp
+++ b/csrc/xpu/lora/lora_ops.cpp
@@ -12,8 +12,9 @@
 //
 //   Tensor shapes:
 //     inputs  : [batch_size, hidden_size]
-//     weights : num_slices tensors, each [num_loras, 1, rank, hidden_size]
-//     output  : [num_slices, batch_size, rank]
+//     weights : num_slices tensors, each [num_loras, 1, stored_rank,
+//               hidden_size]
+//     output  : [num_slices, batch_size, active_rank]
 //     indices : [batch_size]
 //
 // Expand (lora_expand):
@@ -22,8 +23,9 @@
 //         inputs[s, b, :] @ weights_s[indices[b], :, :]
 //
 //   Tensor shapes:
-//     inputs  : [num_slices, batch_size, rank]
-//     weights : num_slices tensors, each [num_loras, 1, slice_size, rank]
+//     inputs  : [num_slices, batch_size, active_rank]
+//     weights : num_slices tensors, each [num_loras, 1, slice_size,
+//               stored_rank]
 //     output  : [batch_size, total_output_dim]
 //     indices : [batch_size]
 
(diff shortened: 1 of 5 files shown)


liu-shaojun:
Closing because this change belongs in the downstream analytics-zoo repository and must be based on its v0.26.0 branch. The validated replacement is analytics-zoo/vllm-xpu-kernels#14.

PR #544, opened 2026-08-25
[GDN] Fix token-indexed conv-state layout in causal_conv1d spec-decode kernel

mayuyuace:
## Problem
XPU GDN spec-decode conv1d kernel writes conv state token-indexed (a full
snapshot per speculative token, one per block-table column), but vLLM's
generic Mamba copy (get_conv_copy_spec) expects a single sliding-window cache
line read by row offset. When num_accepted_tokens > 1 at a cache-block
boundary, the wrong conv state migrates into the next block and corrupts
recurrent updates.

## Impact
Qwen3-Next / Qwen3.5 (GDN) on XPU with MTP: long-context generation collapses
into a narrow token/semantic loop, no runtime error. Larger block size only
delays it.

## Fix
Rewrite conv-state read/write in causal_conv1d_spec_kernel to the
single-cache-line sliding-window convention (column 0 + row offset
num_accepted-1), matching the CUDA/Triton reference and get_conv_copy_spec.
No vLLM change: kernel still takes the 2D cache_indices, conv uses column 0,
ssm keeps all columns (matches get_temporal_copy_spec). SSM and non-spec paths
unchanged.

## Test
Updated GDN spec UTs (tests/gdn_attn/test_causal_conv1d.py, test_gdn_attn.py).
Verify on XPU: pytest tests/gdn_attn/. E2E: Qwen3-Next/Qwen3.5 + MTP no longer
collapses, matches no-boundary control.

## Credit
Root cause was identified and reported by @recutita in vllm-project/vllm#53229
(vLLM-side workaround, not merged). This PR implements that kernel-side fix. Token-indexed
layout was introduced by vllm-xpu-kernels#368 (80372569).

## Model-level verification (XPU, real MTP serving)

Reproducer: recutita's harness (request.sh + compare.py) from #53229 —
3 deterministic recurrence prompts, greedy, seed-fixed, ignore_eos,
3800 generated tokens. Compared block 3200 (crosses the cache-block
boundary) vs block 4000->4032 (no-boundary control), same vLLM / model
/ params; only the xpu-kernels build differs.

Config: Qwen3.5-0.8B, MTP num_speculative_tokens=3, bf16, fp8 KV,
max-model-len 4096, max-num-seqs 1, enforce-eager.

Before fix (unpatched kernel):
  variant 1/2/3: DIFF at seq position ~3201-3203
  => output diverges right at the 3200 boundary (bug reproduced).

After fix (patched kernel, incl. #545 strided state_len fix):
  variant 1/2/3: MATCH, 3800 tokens each, coherent output
  => boundary-crossing output is token-for-token identical to the
     no-boundary control. Fix verified.

### DIFF
--- a/benchmark/benchmark_causal_conv1d.py
+++ b/benchmark/benchmark_causal_conv1d.py
@@ -75,6 +75,7 @@ def estimate_bytes_moved(kwargs):
     ba_per_tok = nk * (2 * nv // nk)
     qkv_per_tok = nk * (2 * hk + hv * nv // nk)
     z_per_tok = nv * hv
+    conv_state_len = kwargs["conv_state"].shape[1]
     # {q,k,v} (=qkv_per_tok) plus b, a (2*nv) intermediates written out.
     inter_per_tok = qkv_per_tok + 2 * nv
 
@@ -85,7 +86,7 @@ def estimate_bytes_moved(kwargs):
     bytes_in = n_tok * (qkvz_per_tok + ba_per_tok) * bpe_in
     bytes_z = n_tok * z_per_tok * bpe_out
     bytes_inter = n_tok * inter_per_tok * bpe_out
-    bytes_conv_state = batch * (width - 1) * qkv_per_tok * bpe_in * 2  # RW
+    bytes_conv_state = batch * conv_state_len * qkv_per_tok * bpe_in * 2
     bytes_weights = qkv_per_tok * width * bpe_in
     # Bias is only read when conv_bias is provided.
     bytes_bias = qkv_per_tok * bpe_in if kwargs["conv_bias"] is not None else 0
--- a/benchmark/benchmark_gdn_attn.py
+++ b/benchmark/benchmark_gdn_attn.py
@@ -261,7 +261,9 @@ def _make_spec_inputs(shape: GdnShape, workload: Workload, dtype):
     projected_states_ba = torch.randn(
         (num_actual_tokens, mixed_ba_size), dtype=dtype, device=DEVICE)
     conv_state = torch.randn(
-        (cache_batch_size, width - 1, mixed_qkv_size),
+        (cache_batch_size,
+         width - 1 + num_spec_tokens,
+         mixed_qkv_size),
         dtype=dtype, device=DEVICE)
     ssm_state = torch.randn(
         (cache_batch_size, num_v_heads // tp_size, head_v_dim, head_k_dim),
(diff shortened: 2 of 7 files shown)


recutita:
@mayuyuace Thank you for crediting me in the PR description. Would you mind also adding the following trailer so my contribution is recorded in the commit history?

`Co-authored-by: Katsumi Takeuchi <contact@recutita.com>`

recutita:
Sorry, our updates crossed. Thank you!

mayuyuace:
> @mayuyuace Thank you for crediting me in the PR description. Would you mind also adding the following trailer so my contribution is recorded in the commit history?
> 
> `Co-authored-by: Katsumi Takeuchi <contact@recutita.com>`

Of course, I've already added it.

recutita:
I think #545 may be a blocker for this PR.
The out-of-range `num_accepted_tokens` case raised by Copilot does not appear reachable through vLLM's normal call path, so as far as I can tell, #544 and #545 together look mergeable.
I would appreciate a review of #545 as well.

mayuyuace:
> I think #545 may be a blocker for this PR. The out-of-range `num_accepted_tokens` case raised by Copilot does not appear reachable through vLLM's normal call path, so as far as I can tell, #544 and #545 together look mergeable. I would appreciate a review of #545 as well.

Oh. I just found and was trying to fix this bug.
Thanks for your effort.
I will test and merge it once test pass.

recutita:
@mayuyuace If repository policy permits, could you use either rebase-and-merge or a merge commit rather than squash for #544? I would appreciate preserving my authorship of the #545 commit in `main`.

mayuyuace:
> @mayuyuace If repository policy permits, could you use either rebase-and-merge or a merge commit rather than squash for #544? I would appreciate preserving my authorship of the #545 commit in `main`.

Sure.

recutita:
Thank you, I really appreciate it!

mayuyuace:
@jikunshang @xinyu-intel Please help review this.

YizhouZ (approved):

xinyu-intel (approved):

mayuyuace:
> Thank you, I really appreciate it!

I found that PR can only be merged in the way "Squash and merge".
image

recutita:
It looks like “Rebase and merge” is enabled for this repository.

xinyu-intel:
@jikunshang rebase and merged is not allowed in this repo, can you help here?

recutita:
Rebase merging is enabled at the repository level, but the ruleset for main restricts allowed_merge_methods to ["squash"], so GitHub only shows “Squash and merge.” Using rebase-and-merge would require changing that ruleset. Sorry for the confusion.

recutita:
If changing the ruleset is not possible, I understand that a squash merge is the only option. In that case, I’d appreciate it if the squashed commit could retain my `Co-authored-by` and `Signed-off-by` trailers.

xinyu-intel:
> If changing the ruleset is not possible, I understand that a squash merge is the only option. In that case, I’d appreciate it if the squashed commit could retain my `Co-authored-by` and `Signed-off-by` trailers.

done, thx for the contribution:)

recutita:
Sorry for requesting a rebase merge when the ruleset did not allow it. I’m very happy with the final result. Thank you for preserving the trailers and merging the fix!

PR #545, opened 2026-08-25
Fix GDN conv state length for strided caches

recutita:

## Purpose

PR #544 changes the XPU GDN speculative-decoding conv state to the
sliding-window convention used by vLLM and the CUDA/Triton implementation.
The updated kernel currently derives the effective state length from the
leading stride:

```cpp
const int state_len = conv_states_stride_0 / conv_elems;
```

This is correct for the contiguous cache tensors used by the kernel tests, but
not for the per-layer views used by vLLM. In an end-to-end Qwen3.5-0.8B run,
the conv cache passed to the kernel had:

```text
shape  = (3067, 6, 6144)
stride = (589824, 6144, 1)
```

The stride-based expression therefore produced a state length of 96 instead
of 6. The kernel then read and wrote beyond the selected layer's state and
corrupted adjacent layers in the shared cache allocation, causing MTP output
to diverge immediately.

This follow-up derives the effective state length from the convolution width
and speculative-token count, matching the CUDA/Triton path:

```cpp
const int state_len = Width - 1 + (num_spec_tokens - 1);
```

It also changes the existing MTP conv test to use a per-layer view into a
shared allocation and verifies that neighboring layers remain unchanged. The
public op now rejects speculative conv states shorter than the effective
sliding window, and the remaining direct spec-decode test and benchmark
callers allocate the enlarged state. The benchmark byte models use the actual
state length as well.

## Test Plan

Run the focused XPU GDN conv test:

```bash
python -m pytest -q -ra -p no:cacheprovider \
  tests/gdn_attn/test_causal_conv1d.py \
  -k test_causal_conv1d_mtp
```

Run the direct conv-to-delta speculative-decoding test:

```bash
python -m pytest -q -ra -p no:cacheprovider \
  tests/gdn_attn/test_gated_delta_rule.py \
  -k test_gated_delta_rule_mtp
```

Run the repository pre-commit suite:

```bash
pre-commit run --all-files
```

After this follow-up is merged into `fix/gdn`, PR #544's main-target CI will
run again and build an updated wheel. Use that artifact to repeat the
Qwen3.5-0.8B vLLM end-to-end comparison with MTP enabled and verify that its
output matches the no-boundary control.

## Test Result

The new strided-cache test was first run against the unmodified PR #544 wheel.
It failed on the first case by detecting writes into adjacent layers:

```text
Running 32 items in this shard
FAILED test_causal_conv1d_mtp[float16-True-silu-True-1-4-128-32-128-16-2-1]
Mismatched elements: 65528 / 13172736 (0.5%)
1 failed, 96 deselected in 2.78s
```

The applicable local source checks passed, including ruff, codespell, isort,
clang-format, SPDX-header checks, filename checks, and import-regex checks.

The fixed-kernel local XPU build and focused tests are in progress. Their
results will be added here when they complete. This PR targets `fix/gdn`, while
the GPU test workflow only runs for pull requests targeting `main`; after this
commit is incorporated, PR #544's main-target CI will also rebuild and test the
updated wheel.

N/A. This fixes the internal state-length calculation without changing the
public API or documented behavior.

## AI assistance

Codex CLI (gpt-5.6-sol xhigh) helped inspect the runtime cache layout, identify
the stride-derived state-length error, and prepare the focused fix and
regression test. During my review, I pointed out an inconsistency in one
revision of the patch before settling on the final version. I reviewed the
resulting diff and ran the reported local checks.

### DIFF
--- a/benchmark/benchmark_causal_conv1d.py
+++ b/benchmark/benchmark_causal_conv1d.py
@@ -75,6 +75,7 @@ def estimate_bytes_moved(kwargs):
     ba_per_tok = nk * (2 * nv // nk)
     qkv_per_tok = nk * (2 * hk + hv * nv // nk)
     z_per_tok = nv * hv
+    conv_state_len = kwargs["conv_state"].shape[1]
     # {q,k,v} (=qkv_per_tok) plus b, a (2*nv) intermediates written out.
     inter_per_tok = qkv_per_tok + 2 * nv
 
@@ -85,7 +86,7 @@ def estimate_bytes_moved(kwargs):
     bytes_in = n_tok * (qkvz_per_tok + ba_per_tok) * bpe_in
     bytes_z = n_tok * z_per_tok * bpe_out
     bytes_inter = n_tok * inter_per_tok * bpe_out
-    bytes_conv_state = batch * (width - 1) * qkv_per_tok * bpe_in * 2  # RW
+    bytes_conv_state = batch * conv_state_len * qkv_per_tok * bpe_in * 2
     bytes_weights = qkv_per_tok * width * bpe_in
     # Bias is only read when conv_bias is provided.
     bytes_bias = qkv_per_tok * bpe_in if kwargs["conv_bias"] is not None else 0
--- a/benchmark/benchmark_gdn_attn.py
+++ b/benchmark/benchmark_gdn_attn.py
@@ -261,7 +261,9 @@ def _make_spec_inputs(shape: GdnShape, workload: Workload, dtype):
     projected_states_ba = torch.randn(
         (num_actual_tokens, mixed_ba_size), dtype=dtype, device=DEVICE)
     conv_state = torch.randn(
-        (cache_batch_size, width - 1, mixed_qkv_size),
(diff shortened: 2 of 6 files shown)


recutita:
@mayuyuace I independently identified the strided-cache state-length issue and the undersized speculative conv-state callers before seeing Copilot's review on #544. This PR fixes both and adds validation at the op boundary. Could you please review it when you have a chance?

recutita:
After #545 was merged, I also confirmed that the updated kernel builds and
passes the focused XPU tests in my environment. I used
`docker.io/intel/llm-scaler-vllm:0.14.0-b8.3.1` with its image-native PyTorch
2.10.0+xpu and IntelLLVM 2025.3.2 stack on one Intel Arc Pro B70:

```text
160 passed, 1256 deselected in 6.51s
```

The passing cases comprise 32 production-like strided conv-cache cases, 64
direct conv-to-delta speculative-decoding cases, and 64 fused GDN attention
cases. The strided cases also verify that the neighboring layers in the shared
cache allocation remain byte-for-byte unchanged.

PR #552, opened 2026-08-26
[GDN] Fix XE2 delta epilogue OOB write with non-contiguous token_indx

CySpiegel:
## Purpose

The XE2 `chunk_gated_delta_rule` epilogue looks up `token_indx` once per chunk and block-writes the chunk's `current_chunk_size` rows at a fixed contiguous stride (`chunk_gated_delta_rule_kernels_xe2.hpp`, O_tensor epilogue). That silently assumes token indices are run-contiguous within every sequence. Mixed spec/non-spec batches hand the non-spec population a shuffled index set, so rows land at wrong global positions: silent corruption of neighbouring rows, or an overrun.

Deterministic 2-invocation repro: any warmup call followed by a `(1,2)`-spec / `(1,3)` mixed case produces a live-row diff at a fixed row; a variant shows ±1-neighbour diffs exactly at the spec token positions (an off-by-one-write signature).

Fix at the interface (`gdn_attn_interface.cpp::gated_delta_rule_non_spec`): when `token_indx` is present, run the XE2 kernel into a compact staging buffer with identity indexing (the layout it assumes), then scatter per row with `index_copy_`. The direct path is unchanged when `token_indx` is absent, so spec-free workloads see no change.

Adds `tests/gdn_attn/test_gdn_attn_mixed.py`: a single mixed invocation must be bitwise identical to chaining the two single-population invocations on cloned states, including with fully shuffled token index sets.

## Test Plan

- `pytest tests/gdn_attn` on BMG hardware (2x Arc B70, oneAPI 2026.0), full scope, on this branch (upstream `main` + this commit).
- The new mixed-batch differential (strict bitwise) matrix.
- End-to-end MTP serving (Qwen3.5-class GDN model, TP=2, `num_speculative_tokens=1`) on our fork carrying the same fix.

## Test Result

- tests/gdn_attn: **1303 passed / 0 failed / 128 skipped (4m32s), identical to the pre-change baseline on the same hardware**
- Mixed-batch differential: **12/12 strict bitwise (mixed invocation == chained single-population invocations, incl. fully shuffled token index sets)**
- End-to-end MTP serving on the fork with this fix: conc-8 mixed-batch canary + HumanEval concurrent evals clean; no device asserts across cold-start cycles.

## Related work

This is the interface-side counterpart of the mixed-batch work already on the fork; it is independent of #551 (conv-state epilogue) and touches only the delta-rule non-spec path.

---
AI-assisted change (Claude Code); reviewed, tested, and submitted by a human.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NQyaRmWY7t6EoVomDQ11fw

### DIFF
--- a/csrc/xpu/gdn_attn/gdn_attn_interface.cpp
+++ b/csrc/xpu/gdn_attn/gdn_attn_interface.cpp
@@ -688,28 +688,58 @@ void gated_delta_rule_non_spec(
 
 #ifdef VLLM_XPU_ENABLE_XE2
   if (num_prefills > 0) {
-    const int* token_indx_ptr =
-        non_spec_token_indx.has_value()
-            ? reinterpret_cast<const int*>(non_spec_token_indx->data_ptr())
-            : nullptr;
-
-    chunk_gated_delta_rule_xe2(
-        queue,
-        core_attn_out_active,
-        q,
-        k,
-        v,
-        b,
-        a,
-        A_log,
-        dt_bias,
-        ssm_state,
-        non_spec_query_start_loc,
-        non_spec_state_indices_tensor,
-        has_initial_state,
-        num_prefills,
-        num_decodes,
-        token_indx_ptr);
+    // The XE2 delta kernel writes core_attn_out in chunk-contiguous blocks
+    // after a SINGLE token_indx lookup per chunk (kernels_xe2.hpp O_tensor
+    // epilogue) — it silently assumes token_indx values are run-contiguous
+    // within every sequence. Mixed spec/non-spec batches produce shuffled
+    // index sets, so rows land at wrong global positions (silent corruption
+    // or overrun). Make the assumption true by construction: with token_indx
+    // present, write into a compact staging buffer with identity indexing,
+    // then scatter per-row.
+    if (non_spec_token_indx.has_value()) {
+      torch::Tensor o_compact = torch::zeros(
+          {non_spec_token,
+           core_attn_out_active.size(1),
+           core_attn_out_active.size(2)},
+          core_attn_out_active.options());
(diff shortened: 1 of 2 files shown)


akapug:
Independent reproduction on the kernels Intel ships in `intel/llm-scaler-vllm:0.26.0-b2` (vllm_xpu_kernels 0.1.11.2.dev0+ga692986), one Arc Pro B70, bf16, using this PR's `test_gdn_attn_mixed.py` with one adaptation: that build rejects a single population over the full batch (`Expected non_spec_token + spec_token == num_actual_tokens`), so the sequential reference runs each population on compactly gathered rows with identity `token_indx` and scatters them back.

I also added a second layout beside the test's `randperm`: each request's tokens contiguous, requests in random order, which is how vLLM builds a batch.

| token layout | batch has a prefill | result |
|---|---|---|
| fully shuffled (as in this PR) | yes | 0 of 8 bitwise identical; relative error 0.76-1.01, 29-48 of 34-51 rows differ |
| each request contiguous | yes | 8 of 8 bitwise identical |
| either | no (decodes and spec decodes) | 8 of 8 bitwise identical |

So the fault reproduces, and it is gross, but on this build only when a request's tokens are not contiguous. That fits what we see in serving (Qwen3.8-27B with MTP on this image, a 41-case task eval at 4 concurrent requests stays 41 of 41). Is there a scheduler or runner path that hands `gdn_attention` a non-contiguous per-request layout? Knowing that would say who is exposed today.

Adapted test and runner: https://github.com/akapug/homelab-atlas/tree/main/b70/vllm-xpu/gdn-mixed-batch

-- Claude Opus 5.5, working in [helm](https://github.com/akapug/helm) for @akapug

PR #557, opened 2026-08-28
[KV offload] Replace Tensor pointer parameters with int64_t to reduce overhead

chaojun-zhang:
## Summary

This PR replaces the `Tensor` parameters in `xpu_host_register` and `xpu_host_unregister` with `int64_t` (pointer address), eliminating unnecessary tensor creation and parsing overhead.

## Motivation

Previously, the functions accepted a `Tensor` containing a single `uint64_t` value that stored the pointer address. This required:

1. Creating a `torch.tensor([ptr], dtype=torch.uint64, device="cpu")` on the Python side
2. Parsing the tensor and extracting the address via `data_ptr<uint64_t>()` on the C++ side

This overhead is unnecessary for a simple address value and adds latency on the hot path.

### C++ functions

**Before:**
```cpp
bool xpu_host_register(const at::Tensor& ptr, int64_t n_bytes);
bool xpu_host_unregister(const at::Tensor& ptr);
```

**After:**
```cpp
bool xpu_host_register(int64_t ptr, int64_t n_bytes);
bool xpu_host_unregister(int64_t ptr);
```

### DIFF
--- a/csrc/ops.h
+++ b/csrc/ops.h
@@ -291,14 +291,9 @@ void xpu_memcpy_sync(
     int64_t kind,
     int64_t device = -1);
 
-// ptr is a single-element uint64 tensor holding the raw address. A scalar
-// int64 op argument cannot losslessly carry every 64-bit pointer bit
-// pattern (addresses with bit 63 set overflow the signed range and fail to
-// cast from Python), so the address is passed via a uint64 Tensor instead,
-// matching the pattern used by swap_blocks_batch.
-bool xpu_host_register(const torch::Tensor& ptr, int64_t n_bytes);
+bool xpu_host_register(int64_t ptr, int64_t n_bytes);
 
-bool xpu_host_unregister(const torch::Tensor& ptr);
+bool xpu_host_unregister(int64_t ptr);
 
 void merge_attn_states(
     torch::Tensor& output,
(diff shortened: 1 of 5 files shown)


mayuyuace (approved):

csrc/ops.h
```diff
@@ -291,14 +291,9 @@ void xpu_memcpy_sync(
     int64_t kind,
     int64_t device = -1);
 
-// ptr is a single-element uint64 tensor holding the raw address. A scalar
```
baodii, line 294: @chaojun-zhang , is this part still right? "addresses with bit 63 set overflow the signed range and fail to cast from Python"

chaojun-zhang: It's safe here through the whole pipeline:
- Python -> PyTorch Op Schema: data_ptr() passes raw 64-bit value as Schema int (int64_t) without overflow.
- C++: static_cast<uintptr_t>(ptr) correctly restores the unsigned address even with bit 63 set.

When passing tensor.data_ptr() into a torch.ops entrypoint, PyTorch C10 layer reinterprets the raw 64-bit bit pattern directly into a C++ int64_t instead of performing range checks, avoiding the OverflowError

baodii (approved):

PR #564, opened 2026-09-02
fix(xpu): preserve tensor strides when pinning non-pinned CPU tensors for UVA view

chaojun-zhang:
## Purpose

When creating an XPU tensor view from an unpinned CPU tensor via `get_xpu_view_from_cpu_tensor`, the existing code called `cpu_tensor.contiguous()` before allocating pinned memory. Calling `contiguous()` flattens non-contiguous strided views (such as transposed weight scales `scale_kn.t()`). Subsequent `.t()` operations on those views produce non-contiguous tensors, triggering unnecessary `.contiguous()` copies during GEMM execution on XPU.

## Changes

- Remove `cpu_tensor.contiguous()` in `get_xpu_view_from_cpu_tensor`.
- Allocate pinned memory directly with `at::empty_like(cpu_tensor, c10::TensorOptions().pinned_memory(true))` and copy via `pinned_owner.copy_(cpu_tensor)`, preserving original shape, layout, and strides.

### DIFF
--- a/csrc/xpu_view.cpp
+++ b/csrc/xpu_view.cpp
@@ -124,7 +124,26 @@ torch::Tensor get_xpu_view_from_cpu_tensor(torch::Tensor& cpu_tensor) {
   auto strides = pinned_owner.strides();
   auto scalar_type = pinned_owner.scalar_type();
 
-  size_t byte_size = pinned_owner.numel() * pinned_owner.element_size();
+  // Compute the number of bytes addressable from `host_ptr` given these
+  // sizes/strides. This must NOT be `numel() * element_size()`: that is
+  // only correct for a "dense" layout (e.g. a plain transpose/permute,
+  // where every element in [0, numel) is used exactly once). A tensor with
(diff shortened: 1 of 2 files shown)


Liangliang-Ma (approved):

mayuyuace (approved):

baodii:
Please make sure https://github.com/chaojun-zhang/vllm-xpu-kernels/blob/9410cc5e800dfa6e7471a9c82d27bb51889e53c5/csrc/xpu_view.cpp#L127 is good for strided tensor. @chaojun-zhang @Liangliang-Ma @mayuyuace

chaojun-zhang:
> Please make sure https://github.com/chaojun-zhang/vllm-xpu-kernels/blob/9410cc5e800dfa6e7471a9c82d27bb51889e53c5/csrc/xpu_view.cpp#L127 is good for strided tensor. @chaojun-zhang @Liangliang-Ma @mayuyuace

updated

PR #570, opened 2026-09-03
[XPU] Add a 2-rank peer-to-peer all-reduce over Level Zero IPC

recutita:
## Purpose

Add an all-reduce kernel for a 2-rank tensor-parallel group that exchanges data through Level Zero IPC and handshakes inside the kernel, plus the IPC interop it needs.

This comes out of vllm-project/vllm#54768 and the RFC in vllm-project/vllm#54766, where the same kernel runs as an opt-in path on the vLLM side. Reviewing that PR, @xinyu-intel asked for the kernels and the Level Zero interop to live here and be exposed to vLLM through Python APIs. This is that move. The vLLM side keeps the communicator, which owns the dma-buf fd exchange, the all-ranks agreement and the routing to oneCCL, and calls the ops registered here.

Earlier revisions of this PR also carried an all-gather and served messages up to 8 MiB, because oneCCL could not be recorded into an XPU graph on this stack. It can: with its SYCL kernels and `CCL_SYCL_ALLREDUCE_SIMPLE_THRESHOLD` and `CCL_SYCL_ALLGATHERV_SIMPLE_THRESHOLD` raised, TP=2 starts and records, and from 64 KiB up oneCCL is level with this kernel or faster. So this revision keeps the all-reduce only, for what vllm-project/vllm#54768 routes here: messages up to 64 KiB.

Measured on 2x Arc Pro B70 (Battlemage, PCIe, no scale-up fabric). A 2-rank bf16 all-reduce, 64 calls recorded into one XPU graph, microseconds per call from launch to completion, two runs of one benchmark:

| | this kernel | oneCCL |
|---|---|---|
| 2 KiB | **6.6, 6.6** | 8.2, 9.4 |
| 10 KiB | **7.2, 7.3** | 9.3, 8.2 |
| 20 KiB | **7.8, 7.8** | 8.4, 8.4 |
| 40 KiB | 9.0, 9.0 | 8.9, 8.9 |
| 64 KiB | 10.7, 10.8 | **9.5, 9.5** |

End to end through vllm-project/vllm#54768 that is +5.2% at one request on Qwen3.5-0.8B and +1.7% on Qwen3.8-27B GPTQ INT4, and level with oneCCL at 64 KiB even though the kernel alone is slower there; see Test Result.

### What the ops are

`xpu_p2p_all_reduce(out, input, my_region, peer_region, slot_bytes)` takes the caller's region and the peer's as raw device addresses, the way vLLM's `custom_all_reduce` passes `fptr_t`. `xpu_p2p_alloc_region(slot_bytes)` returns the region one rank needs for a staging slot of that size: two staging slots, a flag page and a counter page, 136 KiB for vLLM's 64 KiB slot. `xpu_ipc_export_handle` / `open_handle` / `close_handle` / `release_handle` are the Level Zero interop.

`vllm_xpu_kernels/p2p.py` exists for one reason: Level Zero device addresses sit above 2**63 and torch's `int` schema type cannot carry them, so a two's-complement conversion has to happen somewhere. It wraps the region allocation and the IPC helpers and deliberately does not wrap the collective, so a caller converting its pointers once at setup keeps a Python frame out of a path that costs single-digit microseconds.

### Three details that are not obvious from the code

The handshake needs system-scope atomics. An inbound PCIe write from the peer GPU does not invalidate this device's cache, so a `volatile` poll spins on a stale line: measured on 2x B70, it missed a mid-spin arrival after 100 seconds while an atomic poll saw it immediately. The generated SPIR-V was checked to confirm the atomics lower to `CrossDevice` scope.

The kernel carries no per-call state from the host. The sequence number that drives the handshake and picks one of two staging slots lives in a per-workgroup counter in device memory that the kernel advances itself, so a replayed launch keeps making progress with its arguments frozen at capture time. An earlier version passed a host-incremented sequence number as a kernel argument; it froze at capture and silently produced wrong output on replay.

The region is an allocation of its own, taken directly from Level Zero, and the op refuses a region that does not start on a page boundary. On 2x B70 the peer's writes over PCIe disturbed this device's own writes elsewhere in the same allocation: with the region carved out of torch's caching allocator, torch ops on neighbouring tensors returned wrong values now and then, and a slot tail sharing the flag page was corrupted on every full-slot call. The flag page therefore holds nothing this device writes.

### Scope

Two ranks only, and single-tile devices only. Neither is an untested configuration that could be unlocked by removing a check: the kernel exchanges with exactly one peer, and Level Zero has no plural `zeMemGetIpcHandle*`, so a multi-tile allocation cannot be exported as one handle. `reject_multi_tile_allocation` refuses rather than mapping part of a range, and names the `ZE_FLAT_DEVICE_HIERARCHY=FLAT` workaround.

This targets cards without a scale-up fabric, where the collective goes over PCIe. On parts with Xe Link the picture is different and oneCCL is the right answer.

## Test Plan

```bash
python -m pytest tests/test_p2p_collective.py -q
```

Needs two XPUs and Level Zero IPC export. The suite skips with a named reason on a single GPU, and on a multi-tile part under `ZE_FLAT_DEVICE_HIERARCHY=COMPOSITE` where export is refused by design.

With vLLM's 64 KiB staging slot, covers every supported dtype against a reference reduction, sizes from one element to a full slot, 300 back-to-back mixed-size calls, an over-slot input raising, a region not on a page boundary raising, an empty input, cross-device `out` raising, that the device-side counter advances exactly once per launch, and that repeated export and release does not leak file descriptors.

End to end through vllm-project/vllm#54768: Qwen3.5-0.8B (bf16) and Qwen3.8-27B GPTQ INT4, TP=2, XPU graph, `--max-model-len 8192`, `--max-num-seqs 64`, 300 tokens, on `vllm/vllm-openai-xpu:nightly-7c5dc571` with this branch built against the image's torch; both columns use that same build. oneCCL runs with its SYCL kernels and the two thresholds above. One server per model, concurrency and backend, oneCCL first, two runs in each.

## Test Result

2x Intel Arc Pro B70, oneAPI 2026.1, AOT for `bmg`, torch 2.13.0+xpu, 15 consecutive runs, each:

```
11 passed in 2.56s
```

End to end, tok/s, mean of the two runs in each server:

| Model | Concurrent requests | Decode all-reduce | oneCCL | with this kernel | |
|---|---|---|---|---|---|
| Qwen3.5-0.8B | 1 | 2 KiB | 342.38 | 360.21 | **+5.2%** |
| | 4 | 8 KiB | 1224.14 | 1286.75 | **+5.1%** |
| | 16 | 32 KiB | 3803.10 | 3881.86 | **+2.1%** |
| | 32 | 64 KiB | 6046.00 | 6085.15 | +0.6% |
| | 64 | 128 KiB, oneCCL | 8512.49 | 8519.49 | +0.1% |
| Qwen3.8-27B GPTQ INT4 | 1 | 10 KiB | 54.00 | 54.90 | **+1.7%** |
| | 4 | 40 KiB | 194.49 | 197.66 | **+1.6%** |
| | 16 | 160 KiB, oneCCL | 543.15 | 543.08 | 0.0% |
| | 32 | 320 KiB, oneCCL | 527.62 | 527.67 | 0.0% |
| | 64 | 640 KiB, oneCCL | 607.07 | 607.99 | +0.2% |

Rows marked oneCCL decode above 64 KiB, so vllm-project/vllm#54768 sends them to oneCCL in both columns; they are controls.

Output correctness was checked directly rather than inferred from throughput. The oracle is stock `Qwen/Qwen3.5-0.8B`, because it is the only model I have found that is bit-deterministic run to run on XPU: the same prompt at temperature 0 gives identical logprobs across separate processes. `Qwen3.8-27B` does not, even at TP=1 with one sequence in flight, so it cannot be used for this. I have no explanation for the difference.

Over 2000 tokens at TP=2, greedy, this path and oneCCL agree exactly under XPU graph and eager: identical token ids, no numeric difference in any logprob, a maximum absolute delta of 0 across the shared top-5 candidates, and no position where the top-5 membership differs. In an offline batch where one request crosses from oneCCL to this kernel mid-generation, all 49 requests match oneCCL in token ids and logprobs.

I have only two B70s, so nothing here says anything about more than two ranks or about multi-tile parts.

## AI assistance

This PR was developed with Claude Code:

* Opus 5: directed the investigation, designed and ran the measurements, analysed the results, and reviewed the final diff. It also wrote this revision, the reduction to all-reduces up to 64 KiB.
* Fable 5 and Fable 5.1: wrote the communicator and the kernels under that direction.
* Codex CLI with gpt-5.6-luna xhigh and gpt-5.6-sol xhigh: used as adversarial reviewers against the finished diff, not as authors. Their findings changed the design. They caught that the exported Level Zero IPC handle was never released with `zeMemPutIpcHandle`, a leak inherited from the original implementation; that `CHECK_DEVICE` let `out` live on a different XPU than `input`; that a multi-tile allocation would be exported as a single handle covering only part of the range; and that the test suite would fail rather than skip on a machine where the path is deliberately unavailable. Codex CLI also tried an optimisation of the kernel for this revision, which was measured and not adopted (below).
* Human author: chose the problem and the final scope, rejected conclusions the data did not support, directed the experimental method, reviewed the final diff, and ran the linters, the upstream unit tests and the correctness and degradation tests independently.

<details>
<summary>Where the human author changed the direction</summary>

The model reached several wrong conclusions during this work. Each was overturned by the human author rather than by the model:

* "TP=2 cannot help here" (the comparison was against TP=2's own halved-weight baseline, not against a single card).
* "GPU utilisation is low, so the workload is launch-bound" (it is bandwidth bound; a single card reaches 78 to 91% of measured peak).
* "Level Zero IPC is not worth implementing over host staging" (the measurement that supported this was doing both ranks' work on one thread, and was dominated by a `torch.xpu.synchronize()` call).
* "Collectives are only 9% of the regression, so this is not worth building" (profiling put it at roughly 42%).
* "TP=2 is slower than a single card, so it has no value" (speculative decoding supports TP but not necessarily PP, so for a model that does not fit on one card TP is the only option that keeps it).

Three more from the move to vllm-project/vllm-xpu-kernels#570:

* On multi-tile support, the model wanted to ship a guard and not implement the path, on the grounds that it could not be tested here. The human author pushed to implement it instead, which is what turned up the fact that Level Zero has no plural `zeMemGetIpcHandle*` at any spec version, so a multi-handle export cannot be written at all. The guard stayed, but as a documented limit rather than as an excuse.
* The results table was collapsing three orthogonal axes, rank count, execution mode and speculation, into rows. The human author asked for a matrix. Laying it out that way immediately exposed one cell that had never actually been measured, and that cell was the basis of a finding about graphs not helping with MTP. Measured properly, graphs help there by 10.9% and the finding was withdrawn.
* When end-to-end capture kept failing, the model kept working the failing case. The human author asked whether TP=2 eager passed first, which isolated the problem to a layer none of this code touches.

Three more from this revision:

* An optimisation that moved the transfer to peer writes with ESIMD made the kernel 8 to 19% faster alone but only about 0.5% faster end to end. The human author dropped it as not worth its maintenance cost.
* The throughput table was first re-measured with every concurrency in one server. The human author noticed that this was not how the earlier numbers had been taken and had it re-run with one server per concurrency.
* The human author asked for Qwen3.5-0.8B next to Qwen3.8-27B in the table, which is where the gain at low concurrency is largest.

The human author also required that the selected communicator be confirmed from runtime logs rather than inferred from performance, which caught a round of measurements where the code under test was never actually loaded; and, before this revision, decided a two-layer scope by dropping a third, host-shared-memory path that measured no better than oneCCL at 1 MiB.

</details>


recutita:
Two ranks is where this stops today, and I would like it not to be. The check is a capability limit rather than an untested configuration: the kernel exchanges with exactly one peer, so removing it produces wrong results, not slow ones. Extending to four or eight cards is a rewrite rather than a relaxation, and the part I cannot judge from here is the peer reachability matrix once there are more than two.

I would find it worth doing. I do not have the hardware. I am looking into whether I can get access to a machine with more of these, and if that comes through I would be glad to follow up, but I would put the odds at about even. If someone already has four or eight and is willing to try, I would like to hear how it goes.

recutita:
The four findings above are addressed in 2bfbb9e. On the first one I went wider than the line Copilot flagged: the same implementation-defined conversion occurs at 16 sites across both files, including one in the reverse direction that nothing flagged, where `xpu_ipc_open_handle` returned `reinterpret_cast<int64_t>(ptr)`. There are no raw int64 to pointer casts left in either file. The suite still passes 22/22 on 2x Arc Pro B70.

Separately, a useful side effect of CI here: `build-wheel` produces a full wheel against `torch==2.13.0+xpu`, which is exactly what the image I measure on ships.

Until now I was building this extension myself and overlaying it, with `FA2_KERNELS_ENABLED` and `BASIC_KERNELS_ENABLED` off to keep the build to 46 targets rather than 703. That partial build halves MTP throughput at TP=1, where no collective runs at all, so I left speculative figures out of both descriptions rather than publish numbers I could not attribute. The non-speculative numbers were unaffected and matched the stock image to within 0.2%.

With the CI wheel I can install rather than overlay, and redo those measurements properly. Starting that now, and I will update both descriptions with the results.

recutita:
@xinyu-intel an update on both descriptions, which I have rewritten against a single set of measurements taken today.

Everything now comes from the full wheel CI built on this branch, installed rather than overlaid, so the earlier caveat about a partial build is gone and the speculative numbers are back. Two runs per configuration, both values given rather than averaged. Best against best is +69.7% without speculation and +69.5% with MTP.

I also went back over the oneCCL configuration, since it would be easy to accuse these numbers of being measured against a handicapped baseline. Of the three environment variables I had been carrying, only `CCL_ENABLE_SYCL_KERNELS=0` matters: `CCL_ZE_IPC_EXCHANGE` and `CCL_TOPO_P2P_ACCESS` at their defaults give the same throughput. With `CCL_ENABLE_SYCL_KERNELS` at its default of 1, TP=2 does not start here at all, failing with `UR_RESULT_ERROR_DEVICE_LOST`. So the oneCCL figures are from the slower of the two paths, and not by choice. That is stated in both descriptions now.

I may be slow to reply until tomorrow.

recutita:
@jikunshang @xinyu-intel this is ready for review when you have time. CI is green on both PVC and BMG.

Before that, a correction to something both descriptions said, and what I found chasing it.

Both say that with `CCL_ENABLE_SYCL_KERNELS` at its default of 1, TP=2 does not start here. That is true of vLLM, but not of oneCCL. In a bare two-process `torchrun` the default path runs fine on this machine, and it is much faster than the 15% Intel documents:

| 2-rank all-reduce, bf16 | these kernels | oneCCL SK=1 | oneCCL SK=0 |
|---|---|---|---|
| 16 KiB | 9.6 us | 10.3 | 45.4 |
| 64 KiB | 13.2 | 10.0 | 66.5 |
| 128 KiB | 15.3 | 11.8 | 71.7 |
| 1 MiB | 87.2 | 56.3 | 286.2 |
| 8 MiB | 673.8 | 318.2 | 2080.4 |

So the per-call claim in vllm-project/vllm-xpu-kernels#570 holds only against SK=0. Against SK=1 these kernels are level at 16 KiB and slower from 64 KiB up. Both descriptions are edited to say so.

What does not change is graph capture, and not blocking the host. oneCCL still refuses to be recorded (`|CCL_SYCL| sched algorithms do not support sycl_graph recording`) whichever setting it runs under, so `VLLM_XPU_ENABLE_XPU_GRAPH=1` with `--tensor-parallel-size 2` still cannot be combined without this path.

Getting an end-to-end number on the SK=1 path took some finding. Of the five images I tried, TP=2 starts on exactly one:

| image | vLLM | TP=2 at SK=1 |
|---|---|---|
| `vllm/vllm-openai-xpu:nightly-7c5dc571` | 0.28.1rc1.dev199 | `UR_RESULT_ERROR_DEVICE_LOST` |
| built from [xinyu-intel's gist](https://gist.github.com/xinyu-intel/1e2a7f8138a5f3f98a79b40ed7a7c7d5) | 0.28.1rc1.dev273 | `UR_RESULT_ERROR_DEVICE_LOST` |
| `intel/llm-scaler-vllm:0.14.0-b8.3.1` | 0.14.1.dev0 | starts |
| `intel/llm-scaler-vllm:0.21.0-b3.1` | 0.21.1.dev0 | `UR_RESULT_ERROR_OUT_OF_RESOURCES` |
| `intel/llm-scaler-vllm:0.26.0-b1` | 0.26.1.dev0 | `UR_RESULT_ERROR_OUT_OF_RESOURCES` |

Every one of them fails, or succeeds, inside `determine_available_memory`, on the first collective of the memory profiling run. On `0.21.0-b3.1` setting SK=0 does not help either; it then fails with `ZE_RESULT_ERROR_UNKNOWN` out of `zeCommandQueueExecuteCommandLists`, so that image does not do TP=2 here at all. Same machine, same model, float16 throughout because 0.14 rejects bfloat16 for GPTQ.

On that one image, plain oneCCL at TP=2 eager reaches 42.78 and 42.49 tok/s from 31.19 and 31.26 at TP=1, so x1.365. On the upstream nightly these kernels reach 30.96 and 30.83 from 28.48 and 28.47, so x1.085. Their older stack scales better than mine.

I profiled both at TP=2 to find out where that goes, and it is not the collective. Both windows hold exactly 6450 `tensor_model_parallel_all_reduce` calls, and the host cost per call is lower here:

| per call | llm-scaler 0.14, oneCCL | upstream nightly, these kernels |
|---|---|---|
| `tensor_model_parallel_all_reduce` | 81.3 us | 38.3 us |
| collective op | 69.2 us (`c10d::allreduce_`) | 24.9 us (`_xpu_C::xpu_p2p_all_reduce`) |

Two different vLLM versions, but both traces are instrumented the same way and both windows are the same length.

Where the difference actually shows up is GEMM. Both traces contain exactly 15250 `gemm_kernel` launches; they take 3.3 us each there and 46.9 us each on the upstream nightly. Total kernel time over the same 50-token window is 168 ms against 1398 ms.

So the eager gap is a GEMM gap, not a collective one. On the collective itself these kernels are the faster of the two, by roughly a factor of three per call, and that is against an implementation which already skips the clone. Both traces are instrumented the same way, so read that as a ratio rather than as absolute microseconds.

Between the two measurements I trust the end-to-end profile over the microbenchmark, because it is the collective as vLLM actually calls it.

recutita:
One more thing found while trying to run these kernels on the one image where TP=2 starts. They do not work there, and the reason is below the collective: `zeMemOpenIpcHandle` refuses the peer's handle.

| `libze_intel_gpu` | image | opening a peer handle |
|---|---|---|
| 1.15.39122 | `vllm/vllm-openai-xpu:nightly-7c5dc571`, and the image built from xinyu-intel's gist | works both directions |
| 1.15.37833 | `intel/llm-scaler-vllm` 0.21 and 0.26 | `ZE_RESULT_ERROR_INVALID_ARGUMENT` |
| 1.14.37435 | `intel/llm-scaler-vllm:0.14.0-b8.3.1` | works one direction only |

The oldest one is the interesting case. Rank 0 opens rank 1's allocation and rank 1 fails to open rank 0's, in the same run:

```
[0] received fd=19 → OPEN OK
[1] received fd=18 → OPEN FAILED: zeMemOpenIpcHandle(...) failed: 0x78000004
```

No vLLM is involved. It reproduces in about forty lines: allocate on each rank, `zeMemGetIpcHandle`, pass the dma-buf fd over a unix socket with `SCM_RIGHTS`, splice the received fd into the handle, `zeMemOpenIpcHandle`. Both sides `fstat` the received fd successfully and see the same 2 MiB dma-buf, and on both stacks a same-process reopen of a handle succeeds, so it is the cross-process, cross-device import that is being refused. The handle is 64 bytes and its layout looks the same on both runtimes: the fd as a u32 at offset 0, the exporting pid at offset 16.

I have not worked out whether this is something I should be doing differently or a limitation of those runtimes, so I am not claiming which. If there is a documented minimum compute-runtime for `zeMemGetIpcHandle` between two cards, that would settle it and I will put it in the description. I am happy to share the reproduction if it is useful.

Everything measured in this PR is on 1.15.39122.

recutita:
`xpu_p2p_all_gather` now takes `out_my_offset` and `out_peer_offset` instead of deriving both from a rank id, so a caller can split one collective across several launches: chunk c of the two shards lands in two windows of a single output that no rank-derived offset can express. The kernel already took both destinations as byte offsets, so only the host wrapper changes. The exact-size check on `out` becomes a bounds and non-overlap check, by subtraction so an offset near `INT64_MAX` cannot wrap past it. 22 to 26 tests passing on 2x Arc Pro B70.

The vLLM side is in vllm-project/vllm#54768, where a capture past the staging slot is split rather than refused. That lifts the limit this PR's description used to carry: MTP at `--max-num-seqs 64` needs a 15.9 MB logits all-gather and could not be captured before. Both descriptions are updated, with every number re-measured against the wheel CI built here.

recutita:
@jikunshang could someone re-run `run-unit-tests-pvc`? It fails before any test runs, with `zipfile.BadZipFile` on the wheel the same run had just built, while `build-wheel` and `run-unit-tests-bmg` are green on that commit. I cannot re-run it myself.

recutita:
@mayuyuace @YizhouZ when you have the bandwidth, could I ask for a look at this one?

All ten CI jobs are green and the branch is verified end to end on 2x Arc Pro B70 against the wheel CI built here. Qwen3.8-27B GPTQ INT4 at `--max-num-seqs 64`, TP=2 on this path with graph capture against the best TP=1 run: 54.93 tok/s against 32.30 without speculation, 84.92 against 49.71 with MTP.

It adds an all-reduce and an all-gather for a 2-rank group over Level Zero IPC. They exist for graph capture: oneCCL cannot be recorded into an XPU graph on this stack, so TP=2 runs eager while TP=1 does not. These can be recorded, which is where most of that gap comes from. @xinyu-intel asked for them and the Level Zero interop to live here rather than in vllm-project/vllm#54768, so this PR is that move.

TP=2 on oneCCL is 17.59 in the same server, below TP=1. That is at `CCL_ENABLE_SYCL_KERNELS=0`, which is the slower of the two paths, but vLLM fails to start TP=2 on the default of 1 in four of the five images I tried, so 0 is what serves here.

Two ranks and single-tile parts are the scope by design, not a limit more hardware would lift. B70 is the target and the only part I have tested; B60 and B65 should behave the same.

If an approval is more than you can take on now, comments on the numbers or the scope would still help.

recutita:
I found a oneCCL configuration in steveseguin/b70-optimization-lab that records into an XPU graph at TP=2, so the comparison in the description no longer holds.
This PR is still about 1.6% faster end to end at one request; I suspected noise, but so far it looks like a real difference.
Moving to draft to send more of the traffic to oneCCL and optimize. @mayuyuace @YizhouZ no review needed until I mark it ready again.

recutita:
@xinyu-intel @jikunshang @baodii Ready for review again. With oneCCL now recording into an XPU graph, this is reduced to a small all-reduce only: 435 lines of code, 674 with tests (was 730 and 1191). Against the best non-default oneCCL configuration it is still +5.2% at one request on Qwen3.5-0.8B and +1.7% on Qwen3.8-27B.

Issue #453, opened 2026-07-08
[Bug] XPU moe_align_block_size drops every token except index 0


TheBeaninator:
# [Bug] XPU moe_align_block_size drops every token except index 0

## Summary
The native XPU `moe_align_block_size` op writes only the first token index into
`sorted_token_ids`; all other tokens are replaced with the padding sentinel.
Every downstream consumer of these indices (e.g. the triton wna16 fused-MoE
kernels) silently computes on 1/M of the batch, producing fluent-but-wrong
output that degrades as batch size grows.

## Repro (no model needed, ~10 lines)
```python
import torch
import vllm._custom_ops as ops

topk_ids = torch.zeros(4, 1, dtype=torch.int32, device="xpu")  # 4 tokens -> expert 0
block_size, num_experts = 16, 1
max_pad = topk_ids.numel() + num_experts * (block_size - 1)
sorted_ids = torch.empty(max_pad, dtype=torch.int32, device="xpu")
expert_ids = torch.empty(2, dtype=torch.int32, device="xpu")
num_post = torch.empty(1, dtype=torch.int32, device="xpu")

ops.moe_align_block_size(topk_ids, num_experts, block_size,
                         sorted_ids, expert_ids, num_post, None)
print(sorted_ids[:8].cpu().tolist())
```

**Expected:** `[0, 1, 2, 3, 4, 4, 4, 4]` (tokens 0–3, then padding = numel)
**Actual:** `[0, 4, 4, 4, 4, 4, 4, 4]` — tokens 1–3 are gone.

`expert_ids` and `num_tokens_post_padded` come back correct; only the token
scatter is wrong. Behavior is identical for all block sizes tested (16/32/64/128)
and any M ≥ 2.

## Impact
Any quantized-MoE path routed through the triton fused-MoE kernels on XPU is
numerically wrong (observed as garbage generation on
`Qwen3-Coder-30B-A3B-Instruct` int4). The corruption worsens with batch size
(M=1 is nearly correct because token 0 *is* written), which makes it look like
a model/quantization problem rather than an indexing bug.

## Workaround
A pure-torch replacement (stable sort by expert + exclusive-cumsum placement)
in `vllm/model_executor/layers/fused_moe/moe_align_block_size.py` restores
bit-exact behavior; with it (plus a fix for a separate `silu_and_mul` bug filed
alongside) the 30B MoE generates correctly on a Battlemage card.

## Environment
[...]

PR #572, opened 2026-09-03
[Bugfix] Guard moe_align_block_size against negative expert_id (padding sentinel)

### LINKED ISSUE #571: [Bug] moe_align_block_size missing expert_id < 0 guard corrupts/segfaults on padding sentinel (-1) rows
## Description

`moe_align_block_size` in `csrc/moe/moe_align_sum_kernels.cpp` only guards
`expert_id >= num_experts` when reading `topk_ids[i]`. It never checks
`expert_id < 0`, so a negative `topk_id` is used directly as an array index.

vLLM emits `topk_id = -1` intentionally as a padding sentinel for
cudagraph/DP-padding rows when `VLLM_MOE_SKIP_PADDING=1` (the default) —
see `csrc/moe/topk.cpp:236-239` in this repo, which already asserts
`is_pad_row || indices[idx] >= 0`, confirming `-1` is a real, expected input
to this op, not an invalid-input misuse. The first `profile_run()` forward
pass marks **every** row as padding (`vllm/v1/worker/gpu_input_batch.py`
`make_dummy`), so this path is hit on essentially every XPU MoE model on
current `main`.

The negative index reaches two independent unguarded sites:

- `_moe_align_block_size` (L197): guard exists but only checks the upper
  bound (`expert_id >= num_experts`), not `< 0`.
- `_moe_align_block_size_small_batch_expert` (count loop ~L331, sort loop
  ~L386): **no guard at all**, upper or lower bound.
- `_count_and_sort_expert_tokens` (~L427): same upper-bound-only guard as
  above.

With a negative index, `cumsum_buffer[-1]` / the token-count array is
incremented atomically at an out-of-bounds offset, and the sort loop later
writes a token index into `sorted_token_ids` at an out-of-bounds offset. The
size of the out-of-bounds write scales with `topk_ids.numel()`
(`max_num_batched_tokens`), which is why the observable symptom varies:

- Small `max_num_batched_tokens` → the write lands inside a live allocation
  from the caching allocator and silently corrupts adjacent tensors (e.g.
  a nearby layer's weights).
- Larger `max_num_batched_tokens` (e.g. default `8192`) → the write lands
  outside the heap → deterministic segfault on the very first forward pass.

## Reproduction

Any XPU MoE model run through vLLM with default settings
(`VLLM_MOE_SKIP_PADDING=1`, the default) triggers `profile_run()` before the
first real forward, which marks 100% of rows as padding
(`topk_id = -1` for every row). Observed concretely with
`Qwen/Qwen1.5-MoE-A2.7B-Chat-GPTQ-Int4` (GPTQ-Int4 WNA16, TP=1 and TP=2) on
Intel B60, both via the Triton `TritonWNA16Experts` path and the native SYCL
`XPUExpertsWNA16` path (both call into this same kernel).

## Distinct from #453

#453 ("XPU moe_align_block_size drops every token except index 0", closed in
v0.1.11) reproduced with **all-zero, valid, non-negative** `topk_ids`. This
issue is a different defect: the input itself is out of the valid
`[0, num_experts)` range (`-1`), and the kernel performs an out-of-bounds
memory access rather than a logic bug on valid inputs. Verified `origin/main`
still lacks the `expert_id < 0` check that would guard against this.

## Fix

A guard for this exact defect is proposed in #572
(`if (expert_id >= num_experts || expert_id < 0) continue;`, applied at all
four sites, plus the missing guard entirely in the small-batch variant).

Note `csrc/moe/remap_hidden_states.cpp:61,104,224` has the same unguarded
`expert_map[expert_id]` pattern and is likely susceptible to the same class
of bug; not covered here, filing as a separate follow-up concern for
maintainers to prioritize.

afierka-intel:
## Purpose

Fixes vllm-project/vllm-xpu-kernels#571.

`moe_align_block_size` (SYCL) guards `expert_id >= num_experts` but never `expert_id < 0`, at 4 sites (2 have no bound check at all). `-1` is a legitimate input: vLLM's padding-sentinel mechanism (`VLLM_MOE_SKIP_PADDING`, default on) emits it on essentially every XPU MoE forward (see `csrc/moe/topk.cpp:236-239`), and used as a raw index it causes an out-of-bounds write past `sorted_token_ids` — corrupting a neighboring tensor or segfaulting depending on size.

Fixed: `_moe_align_block_size` (L197) and `_count_and_sort_expert_tokens` (L433) — added the missing lower bound. `_moe_align_block_size_small_batch_expert` (L331/L386) — had no bound check at all. Same pattern already existed at L598 (`moe_sum_slot_active`); this applies it consistently. LoRA kernels call into these same functions, so they're fixed too.

Distinct from #453 (closed): that was valid, all-zero `topk_ids`; this is out-of-bounds access from an invalid (`-1`) input, still present on `origin/main`.

Out of scope: `csrc/moe/remap_hidden_states.cpp:61,104,224` has the same unguarded pattern — flagged in the issue, not fixed here.

Root cause verified by code reading. Hardware repro of the resulting corruption/segfault is in the companion vLLM-side PR vllm-project/vllm#55231 — that PR is a temporary stopgap (gates padding-sentinel emission on XPU) meant to be removed once this kernel fix ships in a released wheel; the repro there is evidence for vLLM's dispatch code, not offered here as evidence for this kernel diff.

## Test Plan

```
.venv/bin/python -m pytest tests/test_moe_align_block_size.py -k "expert_map or negative_topk_ids" -v
```

Extended `test_moe_align_block_size_with_expert_map` with a `mask_inactive_experts` case (ported from vLLM CUDA's `f378f79b7c`) and added `test_moe_align_block_size_negative_topk_ids_no_expert_map` for the previously-unguarded `expert_map=None` path.

## Test Result

Not build/run-verified — no SYCL/oneAPI environment available here. Golden-compare confirms valid tokens still route correctly with `-1` sentinels present; it doesn't by itself prove absence of an out-of-bounds write on hardware (needs a sanitizer build). Please run before merging.

None.

### DIFF
--- a/csrc/moe/moe_align_sum_kernels.cpp
+++ b/csrc/moe/moe_align_sum_kernels.cpp
@@ -195,7 +195,7 @@ void _moe_align_block_size(
 
   for (size_t i = tid; i < numel; i += stride) {
     int expert_id = static_cast<int>(topk_ids[i]);
-    if (expert_id >= num_experts) {
+    if (expert_id >= num_experts || expert_id < 0) {
       continue;
     }
     if (has_expert_map) {
@@ -329,6 +329,9 @@ void _moe_align_block_size_small_batch_expert(
   if (local_id_x >= fill_threads) {
     for (size_t i = tid; i < numel; i += stride) {
       int32_t expert_id = static_cast<int32_t>(topk_ids[i]);
+      if (expert_id >= num_experts || expert_id < 0) {
+        continue;
+      }
       if (has_expert_map) {
         expert_id = expert_map[expert_id];
         // filter invalid expert
@@ -384,6 +387,9 @@ void _moe_align_block_size_small_batch_expert(
 
     for (size_t i = tid; i < numel; i += stride) {
       int32_t expert_id = static_cast<int32_t>(topk_ids[i]);
+      if (expert_id >= num_experts || expert_id < 0) {
+        continue;
+      }
       if (has_expert_map) {
         expert_id = expert_map[expert_id];
         // filter invalid expert
@@ -425,7 +431,7 @@ void _count_and_sort_expert_tokens(
 
   for (size_t i = tid; i < numel; i += stride) {
     int32_t expert_id = static_cast<int32_t>(topk_ids[i]);
-    if (expert_id >= num_experts) {
+    if (expert_id >= num_experts || expert_id < 0) {
       continue;
     }
 
--- a/tests/test_moe_align_block_size.py
+++ b/tests/test_moe_align_block_size.py
@@ -250,15 +250,17 @@ def test_moe_align_block_size(m: int, topk: int, num_experts: int,
 @pytest.mark.parametrize("topk", [2, 4])
 @pytest.mark.parametrize("num_experts", [8])
 @pytest.mark.parametrize("block_size", [64])
+@pytest.mark.parametrize("mask_inactive_experts", [False, True])
 def test_moe_align_block_size_with_expert_map(m: int, topk: int,
                                               num_experts: int,
-                                              block_size: int):
-    """Test moe_align_block_size with expert mapping (EP scenario)"""
-    topk_ids = torch.zeros((m, topk), device="xpu", dtype=torch.int32)
-    for i in range(m):
-        experts = torch.randperm(num_experts, device="xpu")[:topk]
-        topk_ids[i] = experts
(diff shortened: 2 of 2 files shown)


jikunshang (approved):

jikunshang:
cc @mayuyuace PTAL

mayuyuace (approved):

PR #578, opened 2026-09-07
[EP] Fix negative expert IDs in XPU MoE remapping

hlin99:
Avoid indexing expert_map with padding IDs before checking whether the route is valid. This prevents an out-of-bounds read for -1 padding routes when expert parallelism is enabled.

EP has accuracy problem by this bug.
1.  the decode output is garbled
2.  same prompt for two times with temperate=0, decode output is different. (even if it's garbled)

### DIFF
--- a/csrc/moe/remap_hidden_states.cpp
+++ b/csrc/moe/remap_hidden_states.cpp
@@ -6,6 +6,15 @@
 namespace vllm {
 namespace moe {
 
+static inline int to_local_expert_id(int global_expert_id,
+                                     const int* expert_map) {
+  if (global_expert_id < 0) {
+    return -1;
+  }
+  return expert_map == nullptr ? global_expert_id
+                               : expert_map[global_expert_id];
+}
+
 class RowsPerExpertCount {
  public:
   RowsPerExpertCount(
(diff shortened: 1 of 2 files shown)


hlin99:
PowerMoE model with TP1/DP2/EP2 as an example, same prompt("The future of artificial intelligence is") x2 times. 


before the fix, it's garbage output and diff each time:
image

after the fix, the output looks good and same each time:
image

jikunshang:
cc @mayuyuace  can you take a look?

mayuyuace:
Can you figure out why vLLM uses -1 as the padding value?

hlin99:
> Can you figure out why vLLM uses -1 as the padding value?

hi @mayuyuace 
this is vllm definition for all the platforms. https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/moe_align_block_size.py#L23

Note: In the case of expert_parallel, moe_align_block_size initially
considers all experts as valid and aligns all tokens appropriately.
Before the function returns it marks the experts_ids that are not in
the current GPU rank as -1 so the MoE matmuls could skip those blocks.
This requires the num_experts input arg to be the num global experts.

mayuyuace (approved):

jikunshang (approved):

PR #579, opened 2026-09-07
[BugFix] Support rms_norm for stacked 2D weights on XPU

### LINKED ISSUE #573: [Bug]: rms_norm ignores a stacked per-row weight — vLLM DFlash _normalize_context_k applies layer 0's k_norm to every draft layer on XPU
# [Bug]: XPU `rms_norm` ignores a stacked per-row weight — vLLM's DFlash `_normalize_context_k` applies layer 0's k_norm to every draft layer on XPU

## Environment

```
vllm-xpu-kernels 0.1.14.1 ; vLLM 0.28.1rc1.dev278+g73029d424 (vllm/vllm-openai-xpu nightly 2026-09-02) ; torch 2.13.0+xpu
Intel Arc Pro B70 (32 GB), Level Zero driver 26.27.39122.11, IGC 2.38.2 ; Ubuntu 24.04, kernel 7.0, xe
```

## Describe the bug

vLLM's DFlash/DFlash2 draft path (`vllm/model_executor/models/qwen3_dflash.py`, git sha 73029d424, `_normalize_context_k`; call-site lines 482 (`self._k_norm_weights = torch.stack(...)`), 560–569 (`_normalize_context_k` → `ops.rms_norm(all_k_normed, all_k, self._k_norm_weights, eps)`)) -> torch.Tensor: 564:        ops.rms_norm( ) normalizes the context K of all draft layers at once by calling `ops.rms_norm(out, all_k, self._k_norm_weights, eps)` with `all_k` shaped `[L, num_ctx, nkv, hd]` and a **stacked weight `[L, hd]`**, with the comment "The weight is selected per layer by the outermost (layer) index." That relies on the CUDA kernel's 2-D weight support (`weight_stride`, row chosen by the outermost index).

The XPU kernel (`csrc/layernorm.cpp`, `rms_norm`) derives `hidden_size = input.size(-1)`, `num_tokens = input.numel() / hidden_size`, only checks `weight->is_contiguous()`, and indexes the weight as a flat `[hidden_size]` array (`weight[idx]` / `v_w[idx]`). So on XPU **every layer's K is normalized with layer 0's k_norm weight**; the remaining L-1 rows of the weight are never read. This is dtype-independent and silently degrades DFlash/DFlash2 draft acceptance on XPU (bf16 included).

## Reproduction (direct kernel test, run on the B70 inside the nightly container)

```python
import torch
from vllm import _custom_ops as ops
dev="xpu"; dt=torch.float16; L,N,nkv,hd=2,4,8,128; eps=1e-6
x=torch.randn(L,N,nkv,hd,device=dev,dtype=dt)
w=torch.ones(L,hd,device=dev,dtype=dt); w[1]=0.0        # layer-1 weight all zeros -> per-row kernel must zero layer 1's output
out=torch.empty_like(x); ops.rms_norm(out,x,w,eps)
ref=x.float()*torch.rsqrt(x.float().pow(2).mean(-1,keepdim=True)+eps)   # = layer-0 (weight 1) result
print((out[0].float()-ref[0]).abs().max())   # 0.0010  layer 0 correct
print(out[1].float().abs().max())            # 3.793   layer 1 NOT zeroed
print((out[1].float()-ref[1]).abs().max())   # 0.0010  layer 1 == result with layer 0's weight
```

Expected (CUDA semantics the call site assumes): `out[1]` is all zeros. Observed on XPU: `out[1]` equals the layer-0-weight result.

## Suggested fix

Either add 2-D weight support to the XPU `rms_norm` (weight stride = hidden_size, row = token_idx / (numel_per_row)), matching `csrc/libtorch_stable/layernorm_kernels.cu` in vLLM, or have the vLLM call site fall back to a per-layer loop when `weight.dim() == 2` on platforms whose kernel lacks it. A shape check that rejects a 2-D weight would at least turn the silent mismatch into an error.

Related: vllm-project/vllm#55250 (DFlash2 fp16 0% acceptance on XPU; separate cause — NaN inside the draft decoder layers — but this bug affects the same path in every dtype).

ccrhx4:
Fix https://github.com/vllm-project/vllm-xpu-kernels/issues/573

XPU's rms_norm kernel does not support stacked 2D weights. When callers pass a stacked 2D weight of shape [num_rows, hidden_size] (one row per layer) to normalize multiple layers in a single kernel launch, every row silently read weight row 0 instead of its own row, producing wrong results with no error.

### How it is fixed:
- In the rms_norm() entry point, derive a weight_stride local variable: 0. The public rms_norm op schema is unchanged.
- Pass weight_stride through call_rms_norm_kernel into all three kernel paths that read weight.
- Each kernel reuses its existing batch_idx (already computed for input/out offsetting) to compute weight_row = weight + batch_idx * weight_stride.
- gemma_rms_norm's call site is updated to pass weight_stride=0 unchanged, since no vLLM caller stacks weights through it today.
- Adds tests/test_batched_weight_rms_norm.py covering batched-weight correctness against a per-row reference loop (5 shapes x 2 dtypes, covering the multi-row, vectorized, and scalar-fallback kernel paths), a direct regression test for the issue's zero-weight-row repro, and shape-validation error tests.

## Test Result

- New tests/test_batched_weight_rms_norm.py (12 parametrized cases): batched-weight correctness vs. a per-row reference loop across 5 shapes x 2 dtypes covering all three kernel paths (multi-row, generic vectorized, scalar fallback for non-power-of-two hidden size), a direct regression test reproducing the issue's zero-weight-row repro, and a shape-validation test for the new TORCH_CHECK errors. All pass.
- Full existing tests/test_layernorm.py (1152 tests) plus tests/test_fused_qk_norm_rope.py, tests/test_fused_input_norm.py, and tests/test_fused_norm_quant.py (808 tests) -- zero regressions on the standard 1D-weight path.
- Cross-validated against vLLM's own tests/kernels/core/test_batched_weight_rms_norm.py via a locally patched (not committed) copy enabling XPU -- 13/13 pass, including float32.
- Verified the issue's exact repro script directly: previously out[1] read layer 0's weight (wrong, e.g. 3.79); now reads its own (zeroed) weight row, giving out[1] == 0.0 as expected.
- Benchmarked with benchmark/benchmark_rmsnorm.py (standard 1D-weight path, both with and without residual): correctness still matches HuggingFace-naive, and vLLM's kernel remains ~9-13x faster than HuggingFace-naive and ~1.2-1.6x faster than torch.compile across the full head_num/batch_size/seq_len sweep -- confirming no performance regression on the common path.

### DIFF
--- a/csrc/layernorm.cpp
+++ b/csrc/layernorm.cpp
@@ -71,7 +71,10 @@ class rms_norm_kernel {
       input_row = input + batch_idx * input_stride_d4 +
                   seq_idx * input_stride_d3 + head_idx * input_stride_d2;
     }
-    const scalar_t* weight_row = weight + batch_idx * weight_stride;
+    const scalar_t* weight_row = nullptr;
+    if constexpr (HasWeight) {
+      weight_row = weight + batch_idx * weight_stride;
+    }
 
     auto vec_op = [&variance](const vec_n_t<scalar_t, VEC_SIZE>& vec) {
 #pragma unroll
@@ -211,7 +214,10 @@ class rms_norm_kernel<scalar_t, NUM_DIMS, 0, HasWeight> {
       input_row = input + batch_idx * input_stride_d4 +
                   seq_idx * input_stride_d3 + head_idx * input_stride_d2;
     }
-    const scalar_t* weight_row = weight + batch_idx * weight_stride;
+    const scalar_t* weight_row = nullptr;
+    if constexpr (HasWeight) {
+      weight_row = weight + batch_idx * weight_stride;
+    }
 
     auto scalar_op = [&variance](const scalar_t& val) {
       float x = static_cast<float>(val);
@@ -344,7 +350,10 @@ class rms_norm_multi_row_kernel {
       input_row = input + batch_idx * input_stride_d4 +
                   seq_idx * input_stride_d3 + head_idx * input_stride_d2;
     }
-    const scalar_t* weight_row = weight + batch_idx * weight_stride;
+    const scalar_t* weight_row = nullptr;
+    if constexpr (HasWeight) {
+      weight_row = weight + batch_idx * weight_stride;
+    }
 
     float variance = 0.0f;
     const int64_t num_vec_elems = hidden_size / VEC_SIZE;
@@ -841,6 +850,9 @@ void rms_norm(
   int64_t weight_stride = 0;
   if (has_weight) {
     TORCH_CHECK(weight->is_contiguous());
+    TORCH_CHECK(
+        weight->scalar_type() == input.scalar_type(),
+        "rms_norm: weight dtype must match input dtype");
     if (weight->dim() == 1) {
       TORCH_CHECK(
           weight->size(0) == input.size(-1),
--- a/csrc/layernorm.cpp
+++ b/csrc/layernorm.cpp
@@ -21,6 +21,7 @@ class rms_norm_kernel {
       const int64_t input_shape_d2_,   // input.size(-2)
       const int64_t input_shape_d3_,   // input.size(-3)
       const scalar_t* weight_,
+      const int64_t weight_stride_,  // 0 for 1D weight, else weight.stride(0)
       const float epsilon_,
       const int num_tokens_,
       const int hidden_size_,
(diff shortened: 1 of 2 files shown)


jikunshang (approved):
thanks for fixing.
cc @zufangzhu

ccrhx4:
I have addressed the copilot review comments. Please help to review. @zufangzhu Thanks a lot.

zufangzhu (approved):
LGTM

PR #580, opened 2026-09-07
[XPU] Fix VEC_SIZE decaying to 1 for act-and-mul kernels at d <= 1024

mganczarenko:

  ## Purpose

  The `VEC_SIZE` heuristic in the act-and-mul kernels (`silu_and_mul`, `gelu_and_mul`, `mul_and_silu`, the FP8 quant variant, `fatrelu_and_mul`, `situ_and_mul`) compares the candidate vector width against
  `tmp_wg = min(d, 1024)`. For `d <= 1024`, `tmp_wg == d`, so the guard is trivially true and `vec_size` always decays to 1 — scalar loads instead of vectorized ones. `d <= 1024` is exactly the per-expert
  intermediate size several real MoE models use, so this silently de-optimizes MoE activation on Intel XPU.

  Fix: also weigh total element count (`num_tokens * d`), not just `d`:

  ```cpp
  const int64_t elems = num_tokens * static_cast<int64_t>(d);
  while (vec_size > 1 &&
         (d % vec_size != 0 ||
          (d / vec_size < 256 && elems < (int64_t(1) << 19)))) {
    vec_size >>= 1;
  }
```
  Applied at all four call sites sharing this heuristic (a fourth site, fatrelu_and_mul, had the identical bug but was missed originally).

  Note: a d-only version of this fix regresses small-batch decode — VEC_SIZE=8 has a real per-call device-time floor once d / vec_size <= 128, invisible to wall-clock but visible in profiler device time.
  The elems term exists specifically to avoid that.

  ## Test Plan

  - tests/test_activation.py unit tests, all four kernels, all dtypes.
  - Correctness vs. a torch reference across the affected d range and several num_tokens.
  - Bandwidth vs. a byte-identical control kernel, to confirm vectorized loads are actually used at d <= 1024.
  - Small-batch sweep across d × num_tokens targeting the decode regime a naive fix would regress.
  - Real MoE model (expert intermediate size in the affected range) end-to-end, prefill/balanced/decode regimes, checking dispatch path and device time.

  ## Test Result

  - Unit tests: 108/108 passed.
  - Correctness: max absolute error 0 at every d/num_tokens.
  - Bandwidth: 1.3–1.75x faster at d<=1024, now matching the control kernel; unchanged at d>=1536.
  - Decode sweep: no regression anywhere (within 1–2% of baseline), including where a naive d-only fix previously regressed 1.30x.
  - Real-model e2e: native CUTLASS MoE dispatch confirmed; fixed kernel measured at 1.8–7.9% of device time depending on regime, coherent output in all cases.

### DIFF
--- a/csrc/activation.cpp
+++ b/csrc/activation.cpp
@@ -432,6 +432,34 @@ class situ_and_mul_kernel {
 
 }  // namespace vllm
 
+namespace {
+
+// Below this many columns per vector lane, VEC_SIZE>1 has a per-call
+// device-time floor that regresses small-batch decode (see PR description).
+constexpr int kActAndMulMinVecCols = 256;
+// Below this total element count, the decode-regime floor above applies;
+// above it, larger batches amortize the floor enough that wider vector
+// loads still win.
+constexpr int64_t kActAndMulMinElemsForNarrowVec = int64_t(1) << 19;
+
+// Shrinks `max_vec_size` until `d` divides evenly by it and either the
+// resulting per-lane column count clears the small-batch decode floor, or
+// the input is large enough to amortize that floor anyway.
+inline int
+ComputeActAndMulVecSize(int d, int64_t num_tokens, int max_vec_size) {
+  const int64_t elems =
+      static_cast<int64_t>(num_tokens) * static_cast<int64_t>(d);
+  int vec_size = max_vec_size;
+  while (vec_size > 1 &&
+         (d % vec_size != 0 || (d / vec_size < kActAndMulMinVecCols &&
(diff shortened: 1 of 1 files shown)


jikunshang:
cc @zufangzhu  PTAL

zufangzhu:
> cc @zufangzhu PTAL


This is exactly the issue I previously fixed in the internal codebase; I will cherry-pick that commit into this public repository. Thank you for your contribution. After I apply that fix, please review the code to see if the issue regarding the `vec` size persists.

jikunshang:
@mganczarenko  thanks for your effort, as #595 landed, let's close this.

PR #582, opened 2026-09-08
add compact_sampling_mask_kernel

mayuyuace:
As this PR show, replace triton kernel by torch reference function can fix CI failure.
https://github.com/vllm-project/vllm/pull/55638
The failure root cause is unknown, and the failure is hard to reproduce by running only the UT case.
But torch reference is slow, so will use sycl kernel to replace triton kernel.
As below results show, sycl kernel is faster than triton kernel.
image

### DIFF
--- a/csrc/xpu/ops.h
+++ b/csrc/xpu/ops.h
@@ -402,3 +402,9 @@ void fused_input_norm(
     torch::Tensor& input,
     torch::Tensor& weight,
     torch::Tensor& bias);
+
+std::tuple<torch::Tensor, torch::Tensor, torch::Tensor> compact_sampling_mask(
+    torch::Tensor& logits,
+    torch::Tensor& num_sampled_tokens,
+    int64_t max_num_kept,
+    int64_t MAX_COMPACT_SUPPORT);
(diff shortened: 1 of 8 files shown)


csrc/xpu/ops.h
```diff
@@ -402,3 +402,9 @@ void fused_input_norm(
     torch::Tensor& input,
     torch::Tensor& weight,
     torch::Tensor& bias);
+
+std::tuple<torch::Tensor, torch::Tensor, torch::Tensor> compact_sampling_mask(
+    const torch::Tensor& logits,
+    const torch::Tensor& num_sampled_tokens,
+    int64_t max_num_kept,
+    int64_t MAX_COMPACT_SUPPORT);
```
jikunshang: emm why use upper case here

jikunshang (approved):

YizhouZ (approved):

csrc/xpu/ops.h
mayuyuace: @copilot+claude-sonnet-5  Help fix the params.

PR #558, opened 2026-08-28
[XE3P] upstream XE3P kernel

YizhouZ:

PR #583, opened 2026-09-08
[XE3P] Upstream Xe3P MoE grouped GEMM and fused MoE interface

Liangliang-Ma:
Bring the Xe3P (CRI / NVL-P) MoE stack over from the internal tree, following the structure of the Xe3P attention upstreaming in #558.

Grouped GEMM:
- Add csrc/xpu/grouped_gemm/xe_3 with its own collective/ headers (array MMA, MXFP and block-FP8 mainloops, epilogue, tile scheduler) and the bf16/fp16/mxfp8/mxfp4/W4A8/block-FP8 policies, plus the prefill tile heuristic.
- Dispatch Xe3P in grouped_gemm_interface.cpp behind VLLM_XPU_ENABLE_XE3P; every other part keeps using the xe_2 kernel.
- Drop the unused xe_default grouped GEMM together with its CMake, setup.py, env and CI plumbing, and remove VLLM_XPU_FORCE_XE_DEFAULT_KERNEL.

The internal tree builds the MXFP mainloop against a cutlass revision that already carries the native/fallback split. The public sycl-tla pin predates it, so the (never dispatched) "unaligned" MXFP policies are left out and the collective is bound to xe_mma_blockscaled_mxfp.hpp.

Fused MoE interface:
- Take weights in the vLLM checkpoint layout [E, N, K] and relayout them for whichever grouped GEMM the device runs, so callers no longer have to pre-transpose. Xe3P additionally gets N-major e8m0 scales.
- Support per-tensor activation scales, static a2_scale, packed mxfp4 activations and grouped-GEMM output dtypes.
- Emit the padded per-expert MN-major activation scale surface directly from remap_hidden_states and the quant kernels via a new expert_scale_desc descriptor, which removes the reorder pass.
- Keep folding activation scales into A on xe_2, which consumes only high-precision activations.

Quantization:
- Add csrc/quantization/moe_scale_layout.h and thread the optional expert_scale_desc through per_token_group_fp8_quant and per_token_group_quant_mxfp4.

Ops:
- Expose is_xe3p_arch, is_cri and is_nvl_p.

Tests:
- Rename the Xe2 MoE tests to *_xe2.py and guard them on BMG/PVC.
- Add test_fused_moe_xe3.py, test_grouped_gemm_xe3.py and test_moe_quant_scale_layout.py.

Validated on BMG (Arc Pro B60): tests/fused_moe/ is 592 passed, 189 skipped (the skips are the Xe3P-only cases). The Xe3P kernels themselves are untested here for lack of hardware.


vllm_xpu_kernels/fused_moe_interface.py
```diff
@@ -198,47 +249,81 @@ def __init__(
         ep_rank=0,
         ep_size=1,
         expert_map=None,
-        gemm1_clamp_limit: Optional[float]=None,
-        activation_situ_beta: Optional[float]=None,
-        activation_situ_linear_beta: Optional[float]=None,
+        is_fp8=None,
```
mayuyuace, line 252: Upstream has remove these is_xxx params.
https://github.com/vllm-project/vllm/pull/48079
Can you also remove these in XE3 kernels?

vllm_xpu_kernels/fused_moe_interface.py
```diff
@@ -198,47 +249,81 @@ def __init__(
...
+        # the Python side below.
+        self._uses_xe2 = _uses_xe2_grouped_gemm(w13)
+
+        # Weights and scales are taken in the layout vLLM loads them in --
+        # [E, N, K] with K packed for 4-bit, plus [E, N, K // group] scales --
+        # and are rewritten in place into whatever this device's grouped GEMM
+        # indexes, so callers never have to know the kernel layouts. The mark
+        # keeps rebuilding over the same parameters a no-op. ref_fused_moe
+        # consumes the loaded layout directly, so it skips the rewrite.
+        if not self._use_ref and not hasattr(w13, "xpu_fused_moe"):
```
mayuyuace, line 283: XpuFusedMoe.__init__() is launched only once, so hasattr(w13, "xpu_fused_moe") is not needed anymore.

mayuyuace:
Over all LGTM.
BTW, do you have accuracy data of each data type models both on XE2 and XE3, and can you please add upstream vllm PR in this PR?

jikunshang (approved):

vllm_xpu_kernels/fused_moe_interface.py
xinyu-intel: here we should make sure w13 not be touched by other class.

PR #597, opened 2026-09-14
build: install psutil in the wheel-build container

SergiioB:
One-line build fix.

`setup.py:compute_num_jobs` imports `psutil` when `MAX_JOBS` is unset, but the in-container dependency list in the docker wheel-build script omits it, so the documented docker wheel-build path fails early with:

```
File "/workspace/vllm-xpu-kernels/setup.py", line 109, in compute_num_jobs
    import psutil
ModuleNotFoundError: No module named 'psutil'
```

`requirements.txt` already carries `psutil`; this adds it to the script's explicit `uv pip install` list.

Verified locally: with this change the docker wheel build passes the dependency stage and proceeds to full compilation (reached the csrc compile stage on Ubuntu 24.04 + oneAPI).


jikunshang (approved):

zufangzhu (approved):

jikunshang:
thanks for fix!

PR #599, opened 2026-09-16
[Bugfix][GDN] Fix gdn_attention handling of ragged n-gram draft lengths

RyanMa29:
## Problem

XPU GDN spec-decode kernels use the configured state-index width as the query length for every request, but vLLM uses `spec_query_start_loc` to describe the actual packed query range. When an n-gram drafter produces fewer tokens than `num_speculative_tokens` (eg, drafter offers 1 token and `num_speculative_tokens=3`), the active query width is smaller than the state-index capacity and the request is rejected by the fixed-width assertion:
https://github.com/vllm-project/vllm-xpu-kernels/blob/1c7cbeee1cd0c1481d48f5031f679a47f3f0ef45/csrc/xpu/gdn_attn/gdn_attn_interface.cpp#L119

## Impact

Qwen3-Next / Qwen3.5 (GDN) on XPU with n-gram speculative decoding: requests with a reduced active speculative width fail with `500 Internal Server Error` and terminate the engine.

## Fix

Use `spec_query_start_loc` in `causal_conv1d_spec_kernel` and `gated_delta_rule_spec_kernel` to determine each request's actual packed query range. Keep the configured state-index width as rollback-state capacity, update only active conv-state and SSM-state slots, and remove the invalid fixed-width token-count assertion. No vLLM change is required. Fixed-width speculative decoding paths are unchanged.

## Test

Added GDN spec-decode coverage for per-request query lengths `[2, 4, 3]`, including packed token remapping and verification that unused SSM rollback slots remain unchanged.

E2E: Qwen3.8-27B-GPTQ-Int4 + n-gram speculative decoding no longer fails when the active draft width is smaller than the configured width.

## Credit

The issue was reported by @jbarrio in https://github.com/vllm-project/vllm/issues/54740. This PR implements the kernel-side fix.

## Model-level verification (XPU, real n-gram serving)

Reproducer: a deterministic completion request whose repeated prompt causes the n-gram proposer to return fewer draft tokens than the configured maximum.

Start vLLM with n-gram speculative decoding:

```bash
vllm serve bernhardbrieger/Qwen3.8-27B-GPTQ-Int4 \
  --quantization gptq \
  --dtype float16 \
  --kv-cache-dtype fp8 \
  --max-model-len 8192 \
  --gpu-memory-utilization 0.92 \
  --max-num-seqs 1 \
  --enable-prefix-caching \
  --speculative-config '{"method":"ngram","num_speculative_tokens":3,"prompt_lookup_min":3,"prompt_lookup_max":3}'
```

Request:

```bash
curl http://127.0.0.1:8000/v1/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "bernhardbrieger/Qwen3.8-27B-GPTQ-Int4",
    "prompt": "intel intel intel",
    "max_tokens": 8,
    "temperature": 0
  }'
```

Before fix (unpatched kernel):

```text
RuntimeError: Expected spec_token == num_spec_decodes * (num_speculative_tokens + 1) to be true, but got false.

POST /v1/completions HTTP/1.1" 500 Internal Server Error
```

After fix (patched kernel):

```text
POST /v1/completions HTTP/1.1" 200 OK
```
The same request completes successfully without the fixed-width assertion. Fix verified.


jikunshang:
please fix dco issue. thanks.

jikunshang:
seems https://github.com/vllm-project/vllm-xpu-kernels/pull/600 introduce some conflicts, please resolve.

RyanMa29:
> seems #600 introduce some conflicts, please resolve.

Thanks. I compared #599 with the merged #600. The kernel fix is functionally equivalent. The additional value in #599 is the n-gram E2E verification, but there is no remaining kernel change to merge after rebasing.

I built the vllm-xpu-kernels with #600, and the n-gram case passed with this update.

yma11:
Close as duplicated PR merged.

